PR #8, opened 2015-05-29
This adds support for using a custom CA cert (or path) using existing Vault env vars

petecheslock:
Vault supports using custom CA certs that can be set for the client
to read
https://github.com/hashicorp/vault/blob/a07bd713ca54f3e33758872b8bc8578627445cdd/command/meta.go#L26-L27

This commit allows the user to use the same ENV vars as they would
using the official vault client in the ruby client.

Thanks for the awesome work on this gem - it's super awesome, can't wait to play with it more.

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -158,6 +158,16 @@ module Vault
           connection.verify_mode = OpenSSL::SSL::VERIFY_PEER
         end
 
+        # Use custom CA cert for verification
+        if ssl_ca_cert
+          connection.ca_file = ssl_ca_cert
+        end
+
+        # Use custom CA path that contains CA certs
+        if ssl_ca_path
+          connection.ca_path = ssl_ca_path
+        end
+
         # Naughty, naughty, naughty! Don't blame when when someone hops in
         # and executes a MITM attack!
         unless ssl_verify
--- a/lib/vault/configurable.rb
+++ b/lib/vault/configurable.rb
@@ -11,6 +11,8 @@ module Vault
         :proxy_port,
         :proxy_username,
         :ssl_pem_file,
+        :ssl_ca_cert,
+        :ssl_ca_path,
         :ssl_verify,
       ]
     end
--- a/lib/vault/defaults.rb
+++ b/lib/vault/defaults.rb
@@ -53,6 +53,19 @@ module Vault
         ENV["VAULT_SSL_CERT"]
       end
 
+      # The path to the CA cert on disk to use for certificate verification
+      # @return [String, nil]
+      def ssl_ca_cert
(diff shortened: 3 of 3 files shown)


sethvargo:
Hiiiiiiiiiiiiiiiiiiiiiiiiii @petecheslock!

Thank you for the PR. Do you think this will cause problems if someone specifies a pem file as well? Maybe we don't care in that case?

petecheslock:
Hiiiiiiiiiiiiiiiii!!!  @sethvargo 
That's actually a good question - also had one other question about those two ENV Vars..

In the vault help
```

  -ca-cert=path           Path to a PEM encoded CA cert file to use to
                          verify the Vault server SSL certificate.

  -ca-path=path           Path to a directory of PEM encoded CA cert files
                          to verify the Vault server SSL certificate. If both
                          -ca-cert and -ca-path are specified, -ca-path is used.
```

So i'm not sure if we want to mirror the way vault works which is checking to see if `ssl_ca_path` is used then use that and ignore `ssl_ca_cert`.  

As for the specifying a pem file - I was trying to figure out in what scenario that would actually be used.  In my use case I have a local internal CA for all my internal stuff.  Then I generate a cert for the vault server (specifically adding SubjectAlternateName stuff which it looks like Vault cares very much about).  From there i'm able to validate the cert on vault since i'm passing the path to my CA cert in my vault request.

Are there any docs about how/where you'd use the ssl pem file thing? I've got a few other things i'm testing today and can try that out and see how it acts.

sethvargo:
@petecheslock I found it from http://augustl.com/blog/2010/ruby_net_http_cheat_sheet/ and http://www.rubyinside.com/nethttp-cheat-sheet-2940.html. They both recommended that approach. _shrugs_. Thoughts?

dkaffee92:
@sethvargo @petecheslock Funny I was about to write the exact same pull request.
@sethvargo, @petecheslock A PEM encoded file can be concatenated to contain both the private key and the certificate which makes these lines valid https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/client.rb#L155-L158

```ruby
          pem = File.read(ssl_pem_file)
          connection.cert = OpenSSL::X509::Certificate.new(pem)
          connection.key = OpenSSL::PKey::RSA.new(pem)
          connection.verify_mode = OpenSSL::SSL::VERIFY_PEER
```

The problem arises in @petecheslock case where the CA that signed the certificate is an internal CA, so you must add the custom `ca_file` / `ca_path` so Ruby will also try to verify the pem against those certificates.
This only adds that file/path to the locations that Ruby will check on the machine to validate that the CA is authorized to sign.

sethvargo:
Merged in e3a42d2. Thanks!

PR #15, opened 2015-07-17
Add a basic integration test for API client read_timeout

BRMatt:
I would add an integration test for ssl timeouts, but they don't seem to work in ruby, my guess is that it's related to problems with `Timeout`.

Hopefully this will do for the meantime, as it tests the timeout that is likely to affect most people (e.g. if a backend vault is using takes too long to respond).

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -208,7 +208,7 @@ module Vault
           error(response)
         end
       end
-    rescue SocketError, Errno::ECONNREFUSED, EOFError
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @BRMatt 

Looks like some of the tests are still failing. Could you please take a look?

BRMatt:
@sethvargo sorry for the delay, tests are passing now.

sethvargo:
Thanks!

PR #16, opened 2015-07-22
Vault requires TLSv1.2 - make sure we use it

james-masson:
Vault is opinionated about using TLSv1.2 - https://github.com/hashicorp/vault/issues/427

The Ruby client may not use it by default, depending on the underlying OpenSSL libraries.
This will typically affect RHEL and similar.

Here's a patch to force the TLS version.


sethvargo:
Merged in 0ac7b74

PR #21, opened 2015-07-27
Implemented AppId class with login function

dylancanfield:
Added a function to authenticate against the App ID backend

### DIFF
--- a/lib/vault/api/app_id.rb
+++ b/lib/vault/api/app_id.rb
@@ -16,13 +16,15 @@ module Vault
     # Create an authentication token.
     #
     # @example
-    #   Vault.app_id.login #=> #<Vault::Secret lease_id="">
+    #   Vault.app_id.login('foo', 'bar') #=> #<Vault::Secret lease_id="">
     #
-    # @param [Hash] options
+    # @param [String] app_id
+    # @param [String] user_id
     #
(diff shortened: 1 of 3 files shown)


lib/vault/api/app_id.rb
```diff
@@ -0,0 +1,31 @@
...
+    # Create an authentication token.
+    #
+    # @example
+    #   Vault.app_id.login('foo', 'bar') #=> #<Vault::Secret lease_id="">
+    #
+    # @param [String] app_id
+    # @param [String] user_id
+    #
+    # @return [Secret]
+    def login(app_id, user_id)
```
sethvargo: I think, as part of this request, we want to save the resulting secret back onto the system in `$HOME/.vault-token` and then also set the Client's .token parameter to the returned value.

sethvargo: Actually, after a bit more consideration, it _shouldn't_ persist to disk, but it should store the resulting token in memory for future requests I think.

dylancanfield:
OK, client's token attribute is now updated for future requests.

sethvargo:
Cleaned up and added more login endpoints in https://github.com/hashicorp/vault-ruby/pull/26. Thanks!

PR #36, opened 2015-08-21
Use a header instead of cookie for the token

sethvargo:


sheldonh:
Against a `0.2.1-dev-41ec36e2a396b305ccc22accd7ff561ad36ad073` server (needed for transitional support of long-lived tokens), I consistently get MissingTokenError without this change. I don't want to impose bundler as a requirement for developers to start using my configuration service.

Any chance of a gem release for this?

Issue #33, opened 2015-08-20
Use github auth method


parabuzzle:
So, maybe I'm blind but with the cli I can do:

```
vault auth -method=github -token=1234...
```

But I don't see a way to use alternate auth methods (github in particular)

sethvargo:
@parabuzzle you can use a few - https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/api/auth.rb, but GitHub isn't currently a first-class one. 

The good news is that auth is pretty easy - it's just an HTTP POST request to the mounted endpoint.

parabuzzle:
I may just add it after work and contrib back then :)

sethvargo:
@parabuzzle thanks :smile: 

For what it's worth, the reason I didn't do it was testing purposes. There's really no good way to integration test it.

parabuzzle:
@sethvargo good question... how strict does integration testing have to be? would it be ok to just stub out the github api calls in the test? (although that may be more work than its worth here)

sethvargo:
Stubs are fine, I just didn't have the time to do it correctly :smile:

parabuzzle:
Ok, I'm stumped... 

I can't find a way to override the Github auth api in the vault server. It seems to only point to `api.github.com`. 

My original thinking was to standup a simple webrick that talks the github api and point the local vault to that for the tests... but without mucking around with DNS/resolver I can't see a way to do that (and I don't want to do that)

Then I thought, why not just create an account and org just for integration testing.. which worked great except Github is too smart for that and invalidated my token when the code was checked in. (Great feature btw). 

So I'm at a loss at this point.. any guidance on this @sethvargo ?

Here's where I have been doing the work: https://github.com/parabuzzle/vault-ruby/tree/add_github_token_support

sethvargo:
@parabuzzle you reached the point where I gave up :smile:. I don't have a good solution at this time until Vault supports custom GH endpoints.

parabuzzle:
damnit.. I need this for an active project.. so I can just use my branch I guess... :(

issue on vault for adding custom GH endpoint support: https://github.com/hashicorp/vault/issues/572

PR #37, opened 2015-08-28
Add GitHub token authentication support

parabuzzle:
Adds the Github token authentication method.

Resolves: https://github.com/hashicorp/vault-ruby/issues/33

The integration tests for this feature rely on the currently unreleased version of Vault. Once the changes that fix https://github.com/hashicorp/vault/issues/572 have been released, the tests for this feature will pass.

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -98,5 +98,27 @@ module Vault
       client.token = secret.auth.client_token
       return secret
     end
+
+    # Authenticate via the Github authentication method. If authentication is
+    # successful, the resulting token will be stored on the client and used
+    # for future requests.
+    #
+    # @example
+    #   Vault.auth.github("mypersonalgithubtoken") #=> #<Vault::Secret lease_id="">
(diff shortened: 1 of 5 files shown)


lib/vault/api/auth.rb
```diff
@@ -98,5 +98,27 @@ def userpass(username, password, options = {})
       client.token = secret.auth.client_token
       return secret
     end
+
+    # Authenticate via the Github authentication method. If authentication is
```
sethvargo: The "H" is capital in GitHub :smile:

parabuzzle: so very true :smile:  I'll fix that later today.

parabuzzle:
is this just not going to be supported or is there another plan for this?

sethvargo:
d40a10611d9da7e60f7aa536717e0b5c2f1bf05f

parabuzzle:
ah.. ok.. nevermind then... I'm an idiot. thanks seth :)

PR #47, opened 2015-12-01
Add support for retries

sethvargo:
This PR adds support for retrying on server (5xx) or socket-level errors that can happen when a Vault server is down or temporarily unavailable.


pearkes:
I'm kind of worried about the retries as a default in a webserver environment, especially with them being used in GET requests – I could see with 2 second timeouts any service with substantial traffic would quickly lockup while servicing requests to an unavailable Vault server.

Is it also possible to send signals/callbacks that retries are being issued that we can use to log or act on otherwise? Think it'll be important for us to know we're doing this.

This implementation looks good to me otherwise, though. Would just recommend we default this to off I think.

lib/vault/defaults.rb
```diff
@@ -75,6 +86,47 @@ def read_timeout
...
+      # {retry_attempts} and {retry_interval} to limit the number of attempts
+      # and the space between them.
+      # @return [Fixnum]
+      def retry_timeout
+        if ENV["VAULT_RETRY_TIMEOUT"].nil?
+          DEFAULT_RETURN_TIMEOUT
+        else
+          ENV["VAULT_RETRY_TIMEOUT"].to_i
+        end
+      end
```
phinze: I think the name of this field is counterintuitive, since there's no "timeout" behavior affected by this value. It's just the maximum time between attempts. `retry_max_sleep` perhaps?

Also there's a constant in the typo, so that should get updated to match the field.

phinze: Haha "typo in the constant", typo-ception!

lib/vault/client.rb
```diff
@@ -325,5 +336,36 @@ def error(response)
...
     end
+
+    # Execute the given block with retries and exponential backoff.
+    def with_retries(rescued = [Exception], &block)
+      exception = nil
+
+      backoff_base = self.retry_base
+      backoff_max  = self.retry_timeout
+
+      self.retry_attempts.times do |attempt|
```
phinze: If I wanted to disable retry behavior, I'd guess setting `retry_attempts` to `0` would do that, but in this case it will cause requests to silently not be made.

Seems like `retry_attempts = 0` should be the thing that makes zero attempts.

lib/vault/client.rb
```diff
@@ -325,5 +336,36 @@ def error(response)
...
+          # Calculate the exponential backoff combined with an element of
+          # randomness.
+          backoff = [backoff_base * (2 ** (attempt - 1)), backoff_max].min
+          backoff = backoff * (0.5 * (1 + Kernel.rand))
+
+          # Ensure we are sleeping at least the minimum interval.
+          backoff = [backoff_base, backoff].max
+
+          # Exponential backoff.
+          sleep(backoff)
```
phinze: Related to above, even "opting out" of retry behavior, I still have to eat this sleep on errors.

lib/vault/client.rb
```diff
@@ -325,5 +336,36 @@ def error(response)
 
       raise HTTPError.new(address, response.code, [response.body])
     end
+
+    # Execute the given block with retries and exponential backoff.
+    def with_retries(rescued = [Exception], &block)
```
phinze: I'd just pull the arg default here since no callers use it. Seems ripe for our far-future selves to accidentally fall back to this behavior when we didn't mean to. :grinning:

phinze:
I too think we should be really careful about back pressure here. I think it's important to be clear to the operator the behavior they're going to get.

Given current implementation, the worst case total time for a Vault request is `timeout * retry_attempts * retry_sleeps` where `retry_sleeps` are defined as the sum of the exponential backoff sleeps starting at `retry_base` and capped at what's currently called `retry_timeout`.

There's also the question of whether we need to be careful about retrying non-idempotent requests.

I wonder if we should start this implementation with a more conservative implementation that affects a smaller percentage of requests.

sethvargo:
> Is it also possible to send signals/callbacks that retries are being issued that we can use to log or act on otherwise? Think it'll be important for us to know we're doing this.

That's a really good idea, but I need some time to think about what that API might look like. Probably some kind of `on_error` handler proc and then the user could log/raise/whatever, but I have to figure out where that should go and how it should bubble up.

> Would just recommend we default this to off I think.

Yea, I thought about that last night. I left the default values for the backoff and interval, but the number of retry attempts is now "0" by default, so it's strictly opt-in behavior.

> There's also the question of whether we need to be careful about retrying non-idempotent requests. I wonder if we should start this implementation with a more conservative implementation that affects a smaller percentage of requests.

We still raise when we get a real response from the server (4xx). To the best of my knowledge, any 500-level request is retryable, which is the only thing we are retrying on.

Cleaned up the other things mentioned.

phinze:
> To the best of my knowledge, any 500-level request is retryable, which is the only thing we are retrying on.

AFAICT these retries also occur on unexpected `EOF` and read timeouts on the socket, which are scenarios in which any request may have completed on the remote side, no?

sethvargo:
@phinze `SocketError` and `EOF` can be raised if the remote server severs the connection unexpectedly (like if Vault were to panic). I think we want to retry in all of those scenarios?

sethvargo:
Refactored this to just provide a high-level function for retries after talking with @phinze

lib/vault/defaults.rb
```diff
@@ -15,6 +15,17 @@ module Defaults
...
+    # The default number of attempts.
+    # @return [Fixnum]
+    RETRY_ATTEMPTS = 2
+
+    # The default backoff interval.
+    # @return [Fixnum]
+    RETRY_BASE = 0.05
+
+    # The default amount of time for a single request to timeout.
+    RETRY_MAX_WAIT = 2.0
```
phinze: This comment should get updated to match the description of the field from the readme, which is nice and clear

pearkes:
I like this approach a lot! Covers our bases and gives us a nice interface for logging/counting retries to circuit break. LGTM.

phinze:
Late to the merge party, but also :+1: :100:

jurriaan:
@sethvargo Awesome! Just a question, why did you choose to roll your own retry logic instead of using a already available solution like [retriable](https://github.com/kamui/retriable)?

README.md
```diff
@@ -78,6 +78,61 @@ client_2 = Vault::Client.new(address: "https://other-vault.mycompany.com")
...
+Vault.with_retries(Vault::HTTPConnectionError, Vault::HTTPError) do |attempt, e|
+  log "Received exception #{e} from Vault - attempt #{attempt}"
+  Vault.logical.read("secret/bacon")
+end
+```
+
+The following options are available:
+
+```ruby
+# :attempts - The number of retries when communicating with the Vault server.
```
jurriaan: Bit confusing, is this the number of retries, or the number of attempts?

PR #48, opened 2015-12-18
Add CGI.escape(path) to logical methods

abooitt:
In case the path contains characters like '%' they should be URI-escaped. This will ensure the same behavior as the command line.

Example:

     vault read 'secret/user@%'

results in this HTTP request:

    GET /v1/secret/user@%25 HTTP/1.1
    Host: 127.0.0.1:8200
    User-Agent: Go-http-client/1.1
    X-Vault-Token: XXXXXX
    Accept-Encoding: gzip

While using the same path with ruby-vault, resulted in the following error:

    bad URI(is not URI?): /v1/secret/user@%

which comes from `/usr/lib/ruby/2.0.0/uri/common.rb:176` 

I considered putting this fix in `lib/vault/client.rb` in the `request` or `build_uri` methods. However, these methods support absolute URIs, which of course should not have the path escaped.

### DIFF
--- a/lib/vault/api/logical.rb
+++ b/lib/vault/api/logical.rb
@@ -1,5 +1,5 @@
-require "cgi"
-
+#require "cgi"
+#
 require_relative "secret"
 require_relative "../client"
 require_relative "../request"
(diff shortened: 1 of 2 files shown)


abooitt:
Thanks!

PR #49, opened 2016-01-06
Add new functions renew_self and revoke_self to the AuthToken Class

jfreimuth:
Add renew_self and revoke_self methods to the AuthToken Class

### DIFF
--- a/spec/integration/api/auth_token_spec.rb
+++ b/spec/integration/api/auth_token_spec.rb
@@ -13,6 +13,14 @@ module Vault
       end
     end
 
+    describe "#renew_self" do
+      it "renew the callign token"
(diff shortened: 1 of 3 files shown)


spec/integration/api/auth_token_spec.rb
```diff
@@ -13,6 +13,14 @@ module Vault
       end
     end
 
+    describe "#renew_self" do
+      it "renew the callign token"
```
sethvargo: Could you write tests for this please?

jfreimuth:
Hi,

i add the spec test, but now you have to update to vault 0.4 for the travis build and then the test will pass. Can you update?

PR #60, opened 2016-02-22
enable configurable retries on all idempotent actions without with_retry block

eddify:
I am hoping this can provide a convenient way of:

- configuring retry options that apply to all my operations
- enable retries without with_retry blocks


sethvargo:
Hi @eddify 

Thank you for the Pull Request. This library is designed to be low-level and expose the required primitives. I do not think the client library should automatically retry since that's ultimately the implementer's decision. The with_retries helper on the vault module also makes this very easy. Thank you though! :smile:

PR #61, opened 2016-03-01
Added ldap authentication

Mike-Petersen:
This will allow to use ldap as authentication backend.

Unfortunately i failed to make tests so far. The service is started and makes a request towards an ldap. In a test this is problematic and i do not know how to mock it here.

Do you have any ideas?

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -98,5 +98,27 @@ module Vault
       client.token = secret.auth.client_token
       return secret
     end
+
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @Mike-Petersen 

Have a look at https://github.com/hashicorp/vault-ruby/pull/37 for an example. You will need to create a fake, in-memory LDAP server most likely.

lizhubertz:
@Mike-Petersen Could you give a code example of how this is meant to be used? There's nothing in the README

PR #72, opened 2016-04-14
Add support for creating tokens using roles

agperson:
If a role option is passed to auth_token.create, use it to construct the endpoint URL (similar to the syntax in the vault CLI).

### DIFF
--- a/lib/vault/api/auth_token.rb
+++ b/lib/vault/api/auth_token.rb
@@ -24,7 +24,12 @@ module Vault
     #
     # @return [Secret]
     def create(options = {})
-      json = client.post("/v1/auth/token/create", JSON.fast_generate(options))
+      url = if options.include?(:role)
+              "/v1/auth/token/create/" + options.delete(:role)
+            else
+              "/v1/auth/token/create"
+            end
+      json = client.post(url, JSON.fast_generate(options))
       return Secret.decode(json)
     end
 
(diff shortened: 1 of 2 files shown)


lib/vault/api/auth_token.rb
```diff
@@ -24,7 +24,12 @@ class AuthToken < Request
     #
     # @return [Secret]
     def create(options = {})
-      json = client.post("/v1/auth/token/create", JSON.fast_generate(options))
+      url = if options.include?(:role)
+              "/v1/auth/token/create/" + options.delete(:role)
+            else
+              "/v1/auth/token/create"
+            end
+      json = client.post(url, JSON.fast_generate(options))
```
Ginja: 👍  on the PR!

Would you need to drop the `:role` key from the `options` hash here too? As it's not a valid parameter. Does the `/v1/auth/token/create/` endpoint simply ignore it?

agperson: The delete method removes the key from the hash and returns the value, so it is no longer present in the options hash.

Verification:
```ruby
irb(main):001:0> h = {test: 'ok', role: 'myrole'}
=> {:test=>"ok", :role=>"myrole"}
irb(main):002:0> h.delete(:role)
=> "myrole"
irb(main):003:0> h
=> {:test=>"ok"}
```

Ginja: Haha, yep. I realized this in the middle of the night. D'oh.

I also would have expected it to have a `!` at the end as it modifies the calling object. Anyways! Would love to see this merged.

PR #75, opened 2016-05-11
Add implementation for the `/sys/mounts/<mount_point>/tune` endpoint.

kvs:
Adds the ability to tune mounts, which also necessitated adding `config` to the `Vault::Mount` struct that gets returned.


sethvargo:
Thanks!

PR #80, opened 2016-07-11
must bump to be greated than 0.4.0

mathieujobin:
Since 0.4.0 already got released.
0.4.0.dev is considered lower than 0.4.0
so if one's app require '>= 0.4.0' and suddenly want to use the latest master.
that requirement won't match

### DIFF
--- a/lib/vault/version.rb
+++ b/lib/vault/version.rb
@@ -1,3 +1,3 @@
 module Vault
-  VERSION = "0.4.0.dev"
+  VERSION = "0.4.1.dev"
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @mathieujobin 

Thank you for opening a Pull Request. This is the desired behavior. While technically "Dev" is higher than non, it's not desirable for users who specified the constraint as `>=` to pick up development versions. Development versions should be explicitly opted into or no version constraint should be specified. Thanks!

PR #81, opened 2016-07-26
Added support for response wrapping and unwrapping.

singram:
Added rudimentary support for response wrapping and unwrapping.

Main focus was to facilitate secure introductions.

Probably needs a clearer way to define the ttl at the user API.  Considered putting it in 'options' but thought I'd solicit feedback first.  The auth_token creation methods could also be dried up depending on final approach.

Fixed minor problem with overriding the token header in client.rb.

### DIFF
--- a/lib/vault/api/logical.rb
+++ b/lib/vault/api/logical.rb
@@ -84,5 +84,42 @@ module Vault
       client.delete("/v1/#{CGI.escape(path)}")
       return true
     end
+
+    # Unwrap the data stored against the given token. If the secret does not exist, +nil+
+    # will be returned.
+    #
+    # @example
+    #   Vault.logical.unwrap("f363dba8-25a7-08c5-430c-00b2367124e6") #=> #<Vault::Secret lease_id="">
+    #
+    # @param [String] wrapper_token
+    #   the token to unwrap
+    #
+    # @return [Secret, nil]
+    def unwrap(wrapper_token)
+      json = client.get("/v1/cubbyhole/response", {}, { Vault::Client::TOKEN_HEADER => wrapper_token })
+      secret = Secret.decode(json)
+      secret.instance_variable_set("@data", Vault::Secret.new(JSON.parse(secret.data[:response], symbolize_names: true))) if secret.data
+      return secret
+    rescue HTTPError => e
+      return nil if e.code == 404
+      raise
+    end
+
+    # Unwrap a token in a wrapped response given the temporary token.
+    #
+    # @example
+    #   Vault.logical.unwrap("f363dba8-25a7-08c5-430c-00b2367124e6") #=> '0f0f40fd-06ce-4af1-61cb-cdc12796f42b'
+    #
+    # @param [String, Secret] wrapper_token
+    #   the token to unwrap as a string or Vault::Secret response
+    #
+    # @return [String]
+    def unwrap_token(wrapper_token)
+      wrapper_token = wrapper_token.wrap_info.token if wrapper_token.is_a?(Vault::Secret) && wrapper_token.wrap_info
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -39,3 +39,5 @@ Gemfile.lock
 
 # Project-specific
 spec/tmp
+/vault
```
sethvargo: Please remove these lines. Thanks!

lib/vault/api/logical.rb
```diff
@@ -40,7 +40,7 @@ def list(path)
     # @param [String] path
     #   the path to read
     #
-    # @return [Secret, nil]
+    # @return [Vault::Secret, nil]
```
sethvargo: Please do not introduce unnecessary changes like this. Thanks!

lib/vault/api/logical.rb
```diff
@@ -84,5 +84,43 @@ def delete(path)
...
+    #
+    # @return [String]
+    def unwrap_token(wrapper_token)
+      wrapper_token = wrapper_token.wrap_info.token if wrapper_token.is_a?(Vault::Secret) && wrapper_token.wrap_info
+      unwrapped_token_response = unwrap(wrapper_token)
+      return unwrapped_token_response.data.auth.client_token
+    rescue HTTPError => e
+      raise
+    end
+
```
sethvargo: Extra newline

lib/vault/client.rb
```diff
@@ -16,6 +16,9 @@ class Client
     # The name of the header used to hold the Vault token.
     TOKEN_HEADER = "X-Vault-Token".freeze
 
+    # The name of the header used to hold the wrapped request ttl.
+    WRAP_TTL_HEADER = "X-Vault-Wrap-Ttl".freeze
```
sethvargo: I think this is actually `X-Vault-Wrap-TTL` (note all caps on TTL)

lib/vault/client.rb
```diff
@@ -132,7 +135,7 @@ def request(verb, path, data = {}, headers = {})
       # Add the Vault token header - users could still override this on a
       # per-request basis
       if !token.nil?
-        request.add_field(TOKEN_HEADER, token)
+        headers[TOKEN_HEADER] ||= token
```
sethvargo: Why change this?

sethvargo:
Hi there,

Thank you for the pull request. I really think `wrap_ttl` should just be an option to the existing options hash instead of introducing a secondary option to the method. This would make the signature:

```ruby
Vault.token.create(wrap_ttl: 50)
```

which seems like a nicer UX to me. What do you think?

There are also some unnecessary changes to the gitignore and some YARD comments. Please revert those. Thanks!

lib/vault/client.rb
singram: Experiments with overriding this header field via the header parameters hash were not working as expected (ruby 2.3.0).  This change fixed that allowing the user to override the value, as per the expectation, as well as consolidating the generation of headers on the request object.

lib/vault/api/logical.rb
singram: Np.  I didn't see a documentation tool in the gemspec but was trying to align the expectations with the other documentation which fully qualify the Secret object.

singram:
Yeah I agree it provides a clearer UX experience for the user.  Any reason not to add yard as part of the developer gemspec so that documentation can be linted etc?

lib/vault/client.rb
singram: Good catch.  Looks like the http proxy I was using for equivalence testing with the go client did some case manipulation.

singram:
@sethvargo All comments should be addressed. Yard was also added as a development dependency and one resulting documentation error fixed.  Let me know if you see anything else that needs addressed.

PR #89, opened 2016-08-26
Fix tests to run against Vault 0.6.1

slyfoxza:
* Bump the version of Vault in the Travis configuration to 0.6.1
* The regex extracting the unseal key from the dev server's startup
  output is changed to match the updated format.
* The userpass and app-id tests associate with the "default" policy
  instead of "root", since root tokens may no longer be created by
  authentication backends.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 
 before_install: |-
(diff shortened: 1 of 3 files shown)


sethvargo:
I just merged a PR that tests a larger matrix to ensure we don't break BC, but thank you for the PR!

PR #90, opened 2016-08-27
bring back silently removed to_h on secret

grosser:
... this used to work on 0.4 ... but then broken when upgrading to 0.5 :(

@sethvargo

### DIFF
--- a/lib/vault/response.rb
+++ b/lib/vault/response.rb
@@ -62,5 +62,11 @@ module Vault
         end
       end
     end
+
+    def to_h
+      self.class.fields.each_with_object({}) do |(k, _), hash|
+        hash[k] = instance_variable_get(:"@#{k}")
+      end
+    end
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @grosser 

Thank you for the PR. This did indeed "break", but it was never an intended public API; this was just a side-effect of having inherited from Struct. Could you explain your use case a bit more? Why do you want to serialize the response into a hash?

grosser:
We have multiple backends for our secret storage and want them to all produce a hash that we then dump in the UI https://github.com/zendesk/samson/blob/master/app/views/admin/secrets/edit.html.erb#L77-L93

sethvargo:
Fixed in master

grosser:
can you add the tests for field / fields too ... not important, but I thought it would be nice to test the basics ...

PR #91, opened 2016-08-27
Test against multiple Vault versions

sethvargo:


legal90:
Hi @sethvargo,
I'm just wondering - how does the Vault version become evaluated during the test? 
I have Vault v0.6.1 installed, but seems like all version-constrained suites are just skipped:

```
$ bundle exec rspec ./spec/integration/api/logical_spec.rb
Run options: exclude {:vault=>#<Proc:./spec/spec_helper.rb:24>}

Randomized with seed 63036

Vault::Logical
  #delete
    allows special characters
    deletes the secret
    does not error if the secret does not exist
  #read
    allows special characters
    returns nil with the thing does not exist
    returns the secret when it exists
  #write
    creates and returns the secret
    overwrites existing secrets
    allows special characters

Finished in 0.58122 seconds (files took 5.58 seconds to load)
9 examples, 0 failures

Randomized with seed 63036
```

Did I miss something?

sethvargo:
Hey @legal90 

https://github.com/hashicorp/vault-ruby/blob/b256513517abc2b31cc177b5378dc9ba2457ae78/spec/spec_helper.rb#L11

legal90:
Oh, thanks! It looks tuned for Travis CI. Ok, I'll handle this on my side.

Issue #88, opened 2016-08-24
vault-ruby incompatible with Vault 0.6.1


SarahKowalik:
We've upgraded to vault 0.6.1, and discovered that vault-ruby doesn't work correctly with it any more.

I've opened up a PR to work with the API data structure changes at https://github.com/hashicorp/vault-ruby/pull/87, but there are another 62 failing tests (https://travis-ci.org/hashicorp/vault-ruby/jobs/154903894).

Do you have an ETA on when vault-ruby will be updated to work with the released Vault server?

Thanks!

grosser:
I'm seeing an issue where the response is no longer parsed as json ... did you see that too ?

ping @sethvargo

SarahKowalik:
Which method?

grosser:
`Vault::Logical#list`

```
TypeError: no implicit conversion of Symbol into Integer
       	from gems/vault-0.5.0/lib/vault/api/logical.rb:29:in `[]'
```

grosser:
getting content type text/plain ... monkeypatch:

```Ruby
# vault returns text/plain and not the expected application/json
# https://github.com/hashicorp/vault-ruby/issues/88
Vault::Client.prepend(Module.new do
  def success(response)
    response.content_type = 'application/json' if response.body && response.body.start_with?('{', '[')
    super
  end
end)
```

jefferai:
Are you using the beta request forwarding support?

sethvargo:
Hi @SarahKowalik,

I fixed up all the Vault 0.6.1 issues and updated the test matrix to be quite verbose to prevent regressions in the future. All the tests are green now.

Hey @grosser,

All the tests are green, and I'm unable to reproduce your issue. Could you please share some additional details and a larger error/stack trace? Are you fronting Vault servers with some kind of LB that could be stripping the content-type? Are you using the new request-forwarding feature that's in beta?

grosser:
issue is https://github.com/hashicorp/vault/issues/1791

grosser:
will be fixed in vault 0.6.2

mfischer-zd:
@sethvargo Can we get a new release with these fixes?

PR #93, opened 2016-08-29
Add to_h to response objects

sethvargo:
Fixes GH-88

### DIFF
--- a/lib/vault/response.rb
+++ b/lib/vault/response.rb
@@ -62,5 +62,28 @@ module Vault
         end
       end
     end
+
+    # Create a hash-bashed representation of this response.
+    #
+    # @return [Hash]
+    def to_h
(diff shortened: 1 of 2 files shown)


lib/vault/response.rb
```diff
@@ -62,5 +62,28 @@ def initialize(opts = {})
...
+
+    # Create a hash-bashed representation of this response.
+    #
+    # @return [Hash]
+    def to_h
+      self.class.fields.inject({}) do |h, (k, opts)|
+        if opts[:as].nil?
+          h[k] = self.public_send(k)
+        else
+          h[k] = self.public_send(opts[:as])
```
grosser: isn't this easier / dryer ? 
`h[k] = self.public_send(opts[:as] || k)`

PR #96, opened 2016-09-16
Adds support for the /sys/auth/<path>/tune API endpoint.

maxvt:


maxvt:
@sethvargo can I please get your eyes on this?

PR #97, opened 2016-09-27
Add support of AppRole auth backend

legal90:
`AppRole` auth backend is available in Vault >= 0.6.1: https://www.vaultproject.io/docs/auth/approle.html

Any notes and suggestions are welcome.

__Caveat:__
Calling `#delete_role` with non-existing role name causes a panic of Vault server. It was fixed by https://github.com/hashicorp/vault/pull/1920 and should work in the upcoming Vault version (>= 0.6.2).

### DIFF
--- a/lib/vault/api/approle.rb
+++ b/lib/vault/api/approle.rb
@@ -1,0 +1,207 @@
+require "json"
+
+require_relative "secret"
+require_relative "../client"
+require_relative "../request"
+require_relative "../response"
+
+module Vault
+  class Client
+    # A proxy to the {AppRole} methods.
+    # @return [AppRole]
+    def approle
+      @approle ||= AppRole.new(self)
+    end
+  end
+
+  class AppRole < Request
+    # Creates a new AppRole or update an existing AppRole with the given name
+    # and attributes.
+    #
+    # @example
+    #   Vault.approle.set_role("testrole", {
+    #     secret_id_ttl: "10m",
+    #     token_ttl:     "20m",
+    #     policies:      "default",
+    #     period:        3600,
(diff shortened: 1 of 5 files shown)


lib/vault/api/approle.rb
```diff
@@ -0,0 +1,207 @@
...
+    # Reads the RoleID of an existing AppRole. If an AppRole does not exist by
+    # that name, +nil+ is returned.
+    #
+    # @example
+    #   Vault.approle.role_id("testrole") #=> #<Vault::Secret lease_id="...">
+    #
+    # @return [Secret, nil]
+    def role_id(name)
+      json = client.get("/v1/auth/approle/role/#{CGI.escape(name)}/role-id")
+      return Secret.decode(json).data[:role_id] || ""
```
sethvargo: What's the situation where this would return nil? Do we want it to return the empty string?

legal90: Ah, sorry. Originally here was `|| nil` (I don't remember why exactly I've switched to `""`).
By the current design, `nil` will be returned in `rescue` section below, when the specified AppRole doesn't exist.
Do you think it would be better to return an empty string in such case?

legal90: Sorry, disregard the previos comment, I've misunderstood the question at first :).
So, yes, there is no reason to return an empty string in this statement, I've removed `|| ""`.
`nil` will be returned only if the specified AppRole doesn't exist, or the `data` doesn't contain a `:role_id` key. But the latter is not an expected behavior for Vault API.

sethvargo: Thanks!

PR #99, opened 2016-10-06
Use Net::HTTP::Persistent

evanphx:
This speeds up using vault by use a persistent HTTP connection

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -1,7 +1,6 @@
 require "cgi"
 require "json"
-require "net/http"
-require "net/https"
+require "net/http/persistent"
 require "uri"
 
 require_relative "configurable"
@@ -53,6 +52,8 @@ module Vault
       # only add them if they are defiend
       a << Net::ReadTimeout if defined?(Net::ReadTimeout)
       a << Net::OpenTimeout if defined?(Net::OpenTimeout)
+
+      a << Net::HTTP::Persistent::Error
     end.freeze
 
     include Vault::Configurable
@@ -177,8 +178,7 @@ module Vault
       # Create the HTTP connection object - since the proxy information defaults
       # to +nil+, we can just pass it to the initializer method instead of doing
       # crazy strange conditionals.
-      connection = Net::HTTP.new(uri.host, uri.port,
-        proxy_address, proxy_port, proxy_username, proxy_password)
+      connection = Net::HTTP::Persistent.new(name: "vault-ruby")
 
       # Use a custom open timeout
       if open_timeout || timeout
@@ -192,8 +192,6 @@ module Vault
 
       # Apply SSL, if applicable
       if uri.scheme == "https"
-        # Turn on SSL
-        connection.use_ssl = true
         connection.verify_mode = OpenSSL::SSL::VERIFY_PEER
 
         # Vault requires TLS1.2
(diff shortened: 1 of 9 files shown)


lib/vault/client.rb
```diff
@@ -177,8 +178,8 @@ def request(verb, path, data = {}, headers = {})
       # Create the HTTP connection object - since the proxy information defaults
       # to +nil+, we can just pass it to the initializer method instead of doing
       # crazy strange conditionals.
-      connection = Net::HTTP.new(uri.host, uri.port,
-        proxy_address, proxy_port, proxy_username, proxy_password)
+      connection = Net::HTTP::Persistent.new(name: "vault-ruby")
```
sethvargo: This seems incorrect. Where are we using the proxy information now? Where is the URI coming from?

lib/vault/client.rb
```diff
@@ -192,9 +193,6 @@ def request(verb, path, data = {}, headers = {})
 
       # Apply SSL, if applicable
       if uri.scheme == "https"
-        # Turn on SSL
-        connection.use_ssl = true
```
sethvargo: Why disable this?

vault.gemspec
```diff
@@ -19,6 +19,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_dependency "net-http-persistent", "~> 3.0"
```
sethvargo: I really really really tried to have zero dependencies in the gem. Could we vendor this or implement it ourselves?

lib/vault/client.rb
evanphx: Net::HTTP::Persistent does it automatically, no need to set that.

lib/vault/client.rb
```diff
@@ -180,6 +180,19 @@ def request(verb, path, data = {}, headers = {})
       # crazy strange conditionals.
       connection = Net::HTTP::Persistent.new(name: "vault-ruby")
 
+      if proxy_address
+        proxy_uri = URI "http://#{proxy_address}"
```
sethvargo: I know it's little, but can we use parens here or URI.parse?

I'm also worried that we are prefixing with http...this would make it impossible to use an HTTPS proxy address.

vault.gemspec
sethvargo: I have had really bad experiences managing gems that are client libraries which consume upstream libs like "restclient", etc. Adding this gem as a requirement means anyone who depends on this gem also must satisfy the same constraint. IMO, client libraries should only ever consume the stdlib for portability.

Especially given the context of vault-rails, it's highly probable to have conflicting constraints on a dependency like net-http-persistent in a large rails app.

lib/vault/client.rb
evanphx: The old code didn't allow for an HTTPS proxy as far as I can tell.

evanphx: No prob.

vault.gemspec
evanphx: This gem is incredibly stable, with the latest release coming this week but previous to that it had been 2 years since the release. I understand your concern about dependencies, but I really do think that using the official gem is better than trying to vendor it.

That being said, if having it as a dependency is a non-starter, I'll work on vendoring.

evanphx:
I vendored net-http-persistent and connection_pool.

lib/vault/client.rb
```diff
@@ -177,8 +179,20 @@ def request(verb, path, data = {}, headers = {})
       # Create the HTTP connection object - since the proxy information defaults
       # to +nil+, we can just pass it to the initializer method instead of doing
       # crazy strange conditionals.
-      connection = Net::HTTP.new(uri.host, uri.port,
-        proxy_address, proxy_port, proxy_username, proxy_password)
+      connection = Net::HTTP::Persistent.new(name: "vault-ruby")
+
+      if proxy_address
+        proxy_uri = URI.parse "http://#{proxy_address}"
```
sethvargo: I don't know where the comment went, but doesn't this eliminate the ability for someone to use an https proxy?

phinze:
`net-http-persistent` is MIT licensed which is compatible with this vendoring, but I think we need to mirror the copyright notice into our license file

PR #101, opened 2016-10-11
allow setting a cert store to use multiple certs or a cert that is no…

grosser:
…t on disk

@evanphx @jonmoter @irwaters


grosser:
@evanphx or @sethvargo merge ?

PR #102, opened 2016-10-11
add privacy and group similar methods

grosser:
@evanphx

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -81,10 +81,41 @@ module Vault
       return nil
     end
 
-    # Determine if the given options are the same as ours.
-    # @return [true, false]
-    def same_options?(opts)
-      options.hash == opts.hash
+    # Execute the given block with retries and exponential backoff.
+    #
+    # @param [Array<Exception>] rescued
+    #   the list of exceptions to rescue
+    def with_retries(*rescued, &block)
+      options      = rescued.last.is_a?(Hash) ? rescued.pop : {}
+      exception    = nil
+      retries      = 0
+
+      max_attempts = options[:attempts] || Defaults::RETRY_ATTEMPTS
+      backoff_base = options[:base]     || Defaults::RETRY_BASE
+      backoff_max  = options[:max_wait] || Defaults::RETRY_MAX_WAIT
+
+      begin
+        return yield retries, exception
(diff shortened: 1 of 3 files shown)


sethvargo:
`with_options?` can't be private. I'm okay making other stuff public, but what's the motivation here?

grosser:
just wanted to clearly mark what is internal ... it is not `with_options?` but `same_options?` ... no idea what this does or who uses it ... could be deleted ?

sethvargo:
Doesn't public vs private already do that? I'm not sure the motivation for marking `same_options?` as private. It's used by "other things", including internal tools. It's just a way to compare if two clients have the same provided option set.

grosser:
that is why I added private ... it did not have anything private before ... 

with 'internal tools' you mean something outside of this repo I assume ?

...

grosser:
restored `same_option?` to public ... still kind of weird that it sits there ...

sethvargo:
Hi @grosser 

I'm still really confused by the motivation for the change? Are you using something and it's not working as expected? The entire API for client is designed to be public, that's why there are no private functions.

grosser:
... I found it very strange that methods like `class_for_request` and `success` are public since they are only useful for internal handling and do not look like they should be exposed ... also they are not tested ... so I'd recommend to make them private to show that they are not supposed to be used and might change at any time ... atm it would mean that non of these methods could be changed without bumping major version since they are all part of the public interface ...

sethvargo:
Hi @grosser 

They are documented and they are meant to be public. They are indirectly tested by all of the other unit tests (i.e. every request uses `class_for_request`. They won't change at any time; they are part of the public API, which is why they are public and documented with YARD.

And yes, you are correct about the version bump, with one caveat. If you read the official [semver spec](http://semver.org/) v2.0.0, you'll note the following two points:

> 4\. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.
> 
> 5\. Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

So, while I have no intentions of doing so, changing a public API pre-1.0 is technically allowed according to the official semantic versioning specification. However, as I just said, I have no intention of doing that and these methods are designed to be public. Thanks and have a great day! 😄

PR #107, opened 2016-10-18
Encode URL paths properly. Fixes #95

### LINKED ISSUE #95: CGI.escape on LDAP groups
Basically what I am running into is I can't use this library to write ldap group policy mappings because CGI.escape subs out spaces and replace them with '+'.  This results in no groups being found for a user.  If I post to this path(auth/ldap/groups/) via the vault CLI things work as expected.  Looking for a work around at the very least.

```irb
irb(main):007:0> client.logical.write("auth/ldap/groups/\"Test Group\"", policies: 'test-policy')
=> true
irb(main):008:0> client.logical.list('auth/ldap/groups')
=> ["\"Test+Group\""]
```

evanphx:
`CGI.escape` uses `+` to escape a ` `, which only valid when used in a query parameter, not in a path. Vault's path decoder treats the `+` as a literal `+`, not a ` `.

This change uses a modified version of `CGI.escape` that encodes ` ` as `%20`.

### DIFF
--- a/lib/vault/encode.rb
+++ b/lib/vault/encode.rb
@@ -1,5 +1,13 @@
 module Vault
   module EncodePath
+
+    # Encodes a string according to the rules for URL paths. This is
+    # used as opposed to CGI.escape because in a URL path, space
+    # needs to be escaped as %20 and CGI.escapes a space as +.
+    #
+    # @param [String]
+    #
+    # @return [String]
     def encode_path(path)
       path.b.gsub(%r!([^a-zA-Z0-9_.-/]+)!) { |m|
         '%' + m.unpack('H2' * m.bytesize).join('%').upcase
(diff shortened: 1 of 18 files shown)


lib/vault/encode.rb
```diff
@@ -0,0 +1,11 @@
+module Vault
+  module EncodePath
+    def encode_path(path)
```
sethvargo: Could we add a detailed comment here explaining:

1. What this does
1. Why it's important

(I know the answers to both of these questions, but I'm thinking about future engineers or community contributors. Bonus points for YARD.)

lib/vault/request.rb
```diff
@@ -37,5 +37,7 @@ def extract_headers!(options = {})
         end
       end
     end
+
+    include EncodePath
```
sethvargo: Nit, but can we put this right below `private`? Or is this a new convention to put modules included privately at the bottom?

sethvargo:
Just two small things. Feel free to merge whenever.

PR #118, opened 2016-11-08
Add list_recursive

grosser:
... POC let me know if you like it and I'll add tests ... using this atm via a monkey-patch since we need to iterate over all keys in multiple places

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -185,6 +185,21 @@ module Vault
       request(:get, path, params, headers)
     end
 
+    # Perform LIST requests to recursively iterate over a directory.
+    # @see Client#request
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @grosser 

Thank you for the Pull Request, but I do not think this is a good idea. For a large data set, this could easily DDOS a Vault cluster without the user even realizing the load it causes. If users want this, I think it's important they understand the algorithm and the operational tradeoffs. Thanks! 😄

grosser:
I'd still prefer having this in vault instead of having multiple devs re-code it ...
can name it `list_recursive_this_is_dangerous` or something funky maybe ...

PR #122, opened 2016-12-13
add aws-ec2 auth backend support

chrisminton:
Add support for the aws-ec2 auth backend.

No rspec tests written 😞 , but I note that the same is true of LDAP, github backends.


sethvargo:
Rebased and merged in https://github.com/hashicorp/vault-ruby/pull/132. Thanks!

PR #135, opened 2017-02-11
Allow aws-ec2 nonce be optional

ostap36:
- Change/fix for the aws-ec2 auth method in order to allow nonce to be an optional parameter from the client's stand point as per the API docs. To support use cases where nonce generation is deferred to the vault backend, rather than having the client supply one.

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -173,11 +173,13 @@ module Vault
     # @param [String] role
     # @param [String] pkcs7
     #   pkcs7 returned by the instance identity document (with line breaks removed)
-    # @param [String] nonce
+    # @param [String] nonce optional
     #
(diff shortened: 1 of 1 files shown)


evanphx:
This looks like really 2 different features? I'd like them separated out into 2 different PRs so they can be consider separately.

ostap36:
Sounds good @evanphx let me update this.

ostap36:
Hi @evanphx I've removed the larger aws-ec2 tests commit, leaving only the nonce optional fix. I'll submit the former in a separate PR. Please let me know if there is anything else, thank you.

ostap36:
@evanphx gentle ping on this one.

ostap36:
Hi @sethvargo does this one require anything else or is it good to merge? Thank you.

PR #136, opened 2017-03-10
Expose pool_size configuration

evanphx:
This allows the connection pool that is used to talk to vault to be configured in size.


evanphx:
@phinze yeah, took me a bit of time to decode that relationship too.

PR #140, opened 2017-04-18
Time.parse requires time

tmaher:
At https://github.com/hashicorp/vault-ruby/blob/447fcb32f9a0331d6856753183219576010d5782/lib/vault/api/secret.rb#L147 , `Time.parse` is called, which necessitates a `require 'time'` ([per the ruby docs](https://ruby-doc.org/stdlib-2.4.0/libdoc/time/rdoc/Time.html)). It's part of stdlib, but you have to explicitly declare it.

You can work around it by doing a `require 'time'` in whatever pulls in the vault gem, but this really should happen in the gem itself.

### DIFF
--- a/lib/vault/api/secret.rb
+++ b/lib/vault/api/secret.rb
@@ -1,4 +1,5 @@
 require_relative "../response"
+require "time"
 
 module Vault
   # Secret is a representation of a secret from Vault. Almost all data returned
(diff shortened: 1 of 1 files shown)


sethvargo:
@tmaher could we move this to the top-level entrypoint instead of this single file? That way we include it at the very end and don't need to worry about including it in future files.

evanphx:
I prefer this PR to moving it, it keeps the dependency close to the usage.

tmaher:
Looking at the other files in that directory, the convention seems to be
```
require "foo"
require "somegem"

PR #142, opened 2017-05-08
Use official SPDX ID for license

kerrizor:
https://spdx.org/licenses/

I'm not sure if this warrants a 0.10.1 release, but maybe?


sethvargo:
Hey @kerrizor 

I think we should release and call this out in bold in the changelog. THanks!

kerrizor:
@sethvargo Took a quick stab at explaining the change, calling out that the licensing scheme has **not changed**.. Edits welcome, naturally.

PR #144, opened 2017-05-31
Add support for aws auth backend for auth_type=iam

shayangz:
Submitting a PR for adding aws iam auth method (released with Vault 0.7.1)

### DIFF
--- a/vault.gemspec
+++ b/vault.gemspec
@@ -19,6 +19,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_runtime_dependency "aws-sdk"
+
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "pry"
   spec.add_development_dependency "rake",    "~> 12.0"
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -1,3 +1,5 @@
+require "aws-sdk"
+require "base64"
 require "json"
 
 require_relative "secret"
@@ -13,6 +15,10 @@ module Vault
   end
 
   class Authenticate < Request
+
+    # canary header used for aws_ec2_iam
+    IAM_SERVER_ID_HEADER = "canaryHeaderValue".freeze
+
     # Authenticate via the "token" authentication method. This authentication
     # method is a bit bizarre because you already have a token, but hey,
     # whatever floats your boat.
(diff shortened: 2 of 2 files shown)


vault.gemspec
```diff
@@ -19,6 +19,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_runtime_dependency "aws-sdk"
```
sethvargo: We keep the client libraries to a zero runtime dependency requirement to make them fast, flexible, and unlikely to conflict with other libraries. Given how heavy the aws-sdk is, I am very opposed to including it as a dependency.

shayangz: Makes sense. 

Do you have a recommendation for how to generate the aws signatures required for the new auth endpoint short of re-implementing a significant amount of logic?

sethvargo: It _should_ just be hmacing a bunch of things.

```ruby
def aws_signature(key, timestamp, region, service)
  k_date    = OpenSSL::HMAC.digest('sha256', "AWS4" + key, timestamp)
  k_region  = OpenSSL::HMAC.digest('sha256', k_date, region)
  k_service = OpenSSL::HMAC.digest('sha256', k_region, service)
  k_signing = OpenSSL::HMAC.digest('sha256', k_service, "aws4_request")
  kSigning
end
```

This looks to be a more complete example: https://gist.github.com/blelump/9e6880a87fefa6397032

sethvargo:
Hey @shayangz 

Did you have a chance to revisit this?

shayangz:
Unfortunately haven't had a chance. It's a bit more involved than just doing  a bunch of HMACs if we want to properly support different AWS auth methods (instance profiles, etc) which I think is necessary for this type of experience.

jhmartin:
Is the mispelling of `iam_auth_header_vaule` intentional?

shayangz:
@jhmartin no, that's a bug 🤦‍♂️.

tristanmorgan:
Hi @shayangz , I created a gist with an attempt to get a vault token via IAM credentials. might give you some hints on how it could work.
https://gist.github.com/tristanmorgan/8fe883f89db4537ce35e0a23f6a537b6

coen-hyde:
AWS has broken up their monolithic aws-sdk gem into a hundred pieces. One of the broken out gems is `aws-sigv4`. Which should be all you need right?

evanphx:
I'd be willing to take a look at the version of this PR that only used the `aws-sigv4` gem.

evanphx:
This happened in #161 instead!

PR #146, opened 2017-07-04
fixes revoke-accessor API path

maciekm:
Getting
```
  * no handler for route 'auth/accessor/revoke-accessor//'
```
Cannot find in the vault changelog when it changed so maybe it never actually worked?


sethvargo:
Thanks! 😄

PR #147, opened 2017-07-13
add yaml configuration file option

ev0ldave:
environment variables take precedence to config file.


sethvargo:
Hi there,

Thank you for the PR. We do not want to support a configuration file for the client library. Sorry!

PR #149, opened 2017-07-19
Add custom mount path support to Authentication#tls

jetpks:
Hi Hashicorp folks!

I'm working on a project that requires several custom cert authentication backends, and I'd really like to use this gem to auth with them. This PR adds a `path` parameter to Authentication#tls to support cert backends mounted at non-default locations.

`rake spec` results: 
Finished in 2.93 seconds (files took 0.58295 seconds to load)
215 examples, 0 failures, 6 pending


sethvargo:
We have to cgi escape the path since it's user-provided.

jetpks:
Good catch. Path is now CGI.escape'd.

sethvargo:
Merged in bd7e60c?

PR #154, opened 2017-11-01
Allow PEM contents as Base64 encoded string

maschwenk:
It is not possible to forward ENV vars with newlines in Docker, thus making it difficult to use `ssl_pem_contents` with Docker. Issue described further here: https://github.com/docker/compose/issues/3527

I realize these can be overridden in initializers/Client config, so would understand if you don't want to directly support this use case and rather allow folks to override it themselves. Will add specs if you give 👍  to implementation.

### DIFF
--- a/lib/vault/defaults.rb
+++ b/lib/vault/defaults.rb
@@ -1,4 +1,5 @@
 require "pathname"
+require "base64"
 
 module Vault
   module Defaults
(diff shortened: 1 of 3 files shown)


evanphx:
This will raise an exception because `ENV.[]` returns nil when the env var is absent. Meaning that if neither variable is defined, an exception is thrown rather than returning nil as is the current behavior. That needs to be dealt with.

maschwenk:
@evanphx I don't think the failures here are due to my changes, looking at master I see the same failure on the last commit

lib/vault/defaults.rb
```diff
@@ -121,12 +122,16 @@ def ssl_ciphers
         ENV["VAULT_SSL_CIPHERS"] || SSL_CIPHERS
       end
 
-      # The raw contents (as a string) for the pem file. To specify the path to
-      # the pem file, use {#ssl_pem_file} instead. This value is preferred over
-      # the value for {#ssl_pem_file}, if set.
+      # The raw contents (as a string) for the pem file, Base64-encoded or unencoded.
```
evanphx: This implies that the output is Base64-encoded? It's always the raw PEM file, i'd leave the comment as it was before.

lib/vault/defaults.rb
```diff
@@ -121,12 +122,16 @@ def ssl_ciphers
...
+      # To specify the path to the pem file, use {#ssl_pem_file} instead.
+      # This value is preferred over the value for {#ssl_pem_file}, if set.
       # @return [String, nil]
       def ssl_pem_contents
-        ENV["VAULT_SSL_PEM_CONTENTS"]
+        if ENV["VAULT_SSL_PEM_CONTENTS_BASE64"]
```
evanphx: Since the new ENV var wins if set, we need to document that somewhere.

maschwenk:
@evanphx Having a tricky time wording that README, open to better suggestions. The difference being that, if you provide it in the configure block, it needs to be the un-encoded string, whereas when passed as an ENV it can be either, prefering base64

PR #161, opened 2017-12-14
Add IAM EC2 auth

maschwenk:
I melded the [gist](https://gist.github.com/tristanmorgan/8fe883f89db4537ce35e0a23f6a537b6) that @tristanmorgan posted and #144. I also used this python [snippet](http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html) on the AWS documentation.

This uses the lightweight `aws-sigv4` gem as suggested by @evanphx. The gist above had a method for achieving it completely without a gem but I noticed the gist had (what appeared to me to be) some errors:

1. It wasn't signing all the headers
2. It wasn't signing the body of the request:

   ```ruby
   k_date    = OpenSSL::HMAC.digest('sha256', 'AWS4' + credentials['SecretAccessKey'], date_stamp)
   k_region  = OpenSSL::HMAC.digest('sha256', k_date, region_name)
   k_service = OpenSSL::HMAC.digest('sha256', k_region, 'iam')
   k_signing = OpenSSL::HMAC.digest('sha256', k_service, 'aws4_request')

   ...

   "SignedHeaders=content-length;content-type;host;x-amz-date;x-vault-server, Signature=#{k_signing}"
   ```

   whereas the ruby AWS SDK [does](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L396) and so does the python snippet:

   ```python
    # Step 7: Combine elements to create create canonical request
    canonical_request = method + '\n' + canonical_uri + '\n' + canonical_querystring + '\n' + canonical_headers + '\n' + signed_headers + '\n' + payload_hash

    algorithm = 'AWS4-HMAC-SHA256'
    credential_scope = datestamp + '/' + region + '/' + service + '/' + 'aws4_request'
    string_to_sign = algorithm + '\n' +  amzdate + '\n' +  credential_scope + '\n' +  hashlib.sha256(canonical_request).hexdigest()
   ```

The basic outline for what I did is this:

1. Use the Instance Metadata API to retrieve the AWS secret key, access key, and region
2. Create an IAM signer for the STS endpoint
3. Sign request with the Vault headers and body included
4. Merge the Sigv4 headers generated by the gem with the Vault headers to create the payload for the Vault API

@evanphx There are a couple open questions on our side that I think you'd be in a good position to answer:

1. Does the Vault role **name** necessarily need to be the same as the AWS role **name**?
2. ~What headers are necessary in `vault_headers`? Do we need things like `content-length`?~ In our testing it worked without it

#### TODO:

* [ ] Tests (I don't see any testing for other auth methods so unsure how to proceed on that)
* [ ] This is a fat method so I'm unsure if you'd like to split it up somehow

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -191,7 +191,7 @@ module Vault
     # for future requests.
     #
     # @example
-    #   Vault.auth.aws_ecs_iam("dev-role-iam", Aws::AssumeRoleCredentials.new, "vault.example.com", "https://sts.us-east-2.amazonaws.com") #=> #<Vault::Secret lease_id="">
+    #   Vault.auth.aws_iam("dev-role-iam", Aws::AssumeRoleCredentials.new, "vault.example.com", "https://sts.us-east-2.amazonaws.com") #=> #<Vault::Secret lease_id="">
     #
     # @param [String] role
     # @param [CredentialProvider] credentials_provider
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -186,7 +186,7 @@ module Vault
       return secret
     end
 
-    # Authenticate via IAM EC2 method by providing a AWS CredentialProvider (either ECS, AssumeRole, etc.)
+    # Authenticate via AWS IAM auth method by providing a AWS CredentialProvider (either ECS, AssumeRole, etc.)
     # If authentication is successful, the resulting token will be stored on the client and used
     # for future requests.
     #
@@ -277,7 +277,7 @@ module Vault
     # Take care changing below regex with that edge case in mind
     #
     # @param [String] sts_endpoint
-    #   The raw pem contents to use for the login procedure.
+    #   https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
     #
     # @return [String] aws region
     def region_from_sts_endpoint(sts_endpoint)
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -197,7 +197,7 @@ module Vault
     # @param [CredentialProvider] credentials_provider
     #   https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CredentialProvider.html
     # @param [String] iam_auth_header_value optional
-    #   As of Jan 2018, Vault will accept ANY or NO header, but this is subject to change and should not be relied upon
+    #   As of Jan 2018, Vault will accept ANY or NO header if none is configured by the Vault server admin
     # @param [String] sts_endpoint optional
     #   https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
     # @return [Secret]
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -13,10 +13,6 @@ module Vault
   end
 
   class Authenticate < Request
-
-    # canary header used for aws_ec2_iam
-    IAM_SERVER_ID_HEADER = "canaryHeaderValue".freeze
-
     # Authenticate via the "token" authentication method. This authentication
     # method is a bit bizarre because you already have a token, but hey,
     # whatever floats your boat.
@@ -203,9 +199,9 @@ module Vault
     # @param [String] sts_endpoint optional
     #   https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
     # @param [String] iam_auth_header_value optional
-    #
+    #   As of Jan 2018, Vault will accept ANY or NO header, but this is subject to change and should not be relied upon
     # @return [Secret]
-    def aws_iam(role, credentials_provider, sts_endpoint = 'https://sts.amazonaws.com', iam_auth_header_value = IAM_SERVER_ID_HEADER)
+    def aws_iam(role, credentials_provider, sts_endpoint = 'https://sts.amazonaws.com', iam_auth_header_value = nil)
       require "aws-sigv4"
       require "base64"
 
@@ -218,9 +214,10 @@ module Vault
 
       vault_headers = {
         'User-Agent' => Vault::Client::USER_AGENT,
-        'Content-Type' => 'application/x-www-form-urlencoded; charset=utf-8',
-        'X-Vault-AWS-IAM-Server-ID' => iam_auth_header_value
+        'Content-Type' => 'application/x-www-form-urlencoded; charset=utf-8'
       }
+
+      vault_headers['X-Vault-AWS-IAM-Server-ID'] = iam_auth_header_value if iam_auth_header_value
 
       sig4_headers = Aws::Sigv4::Signer.new(
         service: 'sts',
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -1,3 +1,5 @@
+require "aws-sigv4"
+require "base64"
 require "json"
 
 require_relative "secret"
@@ -13,6 +15,10 @@ module Vault
   end
 
   class Authenticate < Request
+
+    # canary header used for aws_ec2_iam
+    IAM_SERVER_ID_HEADER = "canaryHeaderValue".freeze
+
     # Authenticate via the "token" authentication method. This authentication
     # method is a bit bizarre because you already have a token, but hey,
     # whatever floats your boat.
@@ -181,6 +187,67 @@ module Vault
       # Set a custom nonce if client is providing one
       payload[:nonce] = nonce if nonce
       json = client.post('/v1/auth/aws-ec2/login', JSON.fast_generate(payload))
+      secret = Secret.decode(json)
+      client.token = secret.auth.client_token
+      return secret
+    end
+
+    # Authenticate via the AWS EC2 authentication method (IAM method). If authentication is
+    # successful, the resulting token will be stored on the client and used
+    # for future requests.
+    #
+    # @example
+    #   Vault.auth.aws_ec2_iam("dev-role-iam", "vault.example.com") #=> #<Vault::Secret lease_id="">
+    #
+    # @param [String] role
+    # @param [String] iam_auth_header_vaule
+    #
+    # @return [Secret]
+    def aws_ec2_iam(role, iam_auth_header_value)
+      aws_meta_data_host = 'http://169.254.169.254'
+      document_uri = URI.join(aws_meta_data_host, '/latest/dynamic/instance-identity/document')
+      document_api_response = Net::HTTP.get(document_uri)
+      document = JSON.parse(document_api_response)
+
+      role_base_uri = URI.join(aws_meta_data_host, '/latest/meta-data/iam/security-credentials/')
+      aws_role_name = Net::HTTP.get(role_base_uri)
+
+      credentials_uri = URI.join(aws_meta_data_host, role_base_uri, aws_role_name)
+      credentials_api_response = Net::HTTP.get(credentials_uri)
+      credentials = JSON.parse(credentials_api_response)
+
+      request_body = 'Action=GetCallerIdentity&Version=2011-06-15'
+      request_url = 'https://sts.amazonaws.com/'
+      request_method = 'POST'
+      iam_auth_header_value ||= IAM_SERVER_ID_HEADER
+
+      vault_headers = {
+        'User-Agent' => Vault::Client::USER_AGENT,
+        'Content-Type' => 'application/x-www-form-urlencoded; charset=utf-8',
+        'X-Vault-AWSIAM-Server-Id' => iam_auth_header_value
+      }
+
+      sig4_headers = Aws::Sigv4::Signer.new(
+        service: 'iam',
+        region: document['region'],
+        access_key_id: credentials['AccessKeyId'],
+        secret_access_key: credentials['SecretAccessKey']
+      ).sign_request(
+        http_method: request_method,
+        url: request_url,
+        headers: vault_headers,
+        body: request_body
+      ).headers
+
+      payload = {
+        role: role,
+        iam_http_request_method: request_method,
+        iam_request_url: Base64.strict_encode64(request_url),
+        iam_request_headers: Base64.strict_encode64(vault_headers.merge(sig4_headers).to_json),
+        iam_request_body: Base64.strict_encode64(request_body)
+      }
+
+      json = client.post('/v1/auth/aws/login', JSON.fast_generate(payload))
       secret = Secret.decode(json)
       client.token = secret.auth.client_token
       return secret
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -205,6 +205,8 @@ module Vault
       require "aws-sigv4"
       require "base64"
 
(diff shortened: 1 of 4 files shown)


maschwenk:
@evanphx What we're working on now is how this can work for both standard EC2 instances versus ECS tasks. ECS tasks use a different credentials API at:
`"http://169.254.170.2#{ENV['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI']}"`

maschwenk:
60f70e2 breaks the ECS/EC2 auth into two separate public methods that share an interface for iam

joelthompson:
@maschwenk -- was thinking about this a bit. The Vault team understandably doesn't want to pull the entirety of the AWS SDK for Ruby into the Vault Ruby API client because a number of users probably won't even want it, and it could cause version conflicts with clients using a different version. And unfortunately, all the "magic" that AWS does on your behalf to figure out where to load credentials from isn't easily separable from the rest of the SDK. And the way that you're doing it won't allow clients to use explicit credentials, environment-variable-provided credentials, or AssumeRole-provided credentials (a workaround I suggest on the mailing list from time to time because I've been too lazy implement allowing binding multiple IAM principal ARNs to a single Vault role); doing any of those would require more code changes.

So, why not just move that little bit of complexity to the client, removing from client code the complexity of figuring out whether they have to call `aws_ec2_iam` or `aws_ecs_iam`? Have clients pass in a [CredentialProvider](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CredentialProvider.html) object. Since Ruby is duck-typed, you don't need to pull in the SDK. If clients want the default credential chain, they just pass in `Aws::CredentialProviderChain.new.resolve` to get the default chain that the SDK uses (and that people are familiar with). If they want AssumeRole-supplied credentials, well, they just do exactly what [the offical AWS documentation](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-credentials-access-token) tells people to do and pass in the results of `Aws::AssumeRoleCredentials.new`

joelthompson:
Also, @maschwenk, for future reference:

> ~~What headers are necessary in vault_headers? Do we need things like content-length?~~ In our testing it worked without it

Per the [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) (see step 5), "The host header must be included as a signed header. If you include a date or x-amz-date header, you must also include that header in the list of signed headers."

maschwenk:
@joelthompson sounds good. The only remaining piece in that puzzle is the region. Whereas previously we had it coming from the AWS Metadata API, we'd now need it to be statically passed in initialization, is that ok?

joelthompson:
@maschwenk:

> The only remaining piece in that puzzle is the region. Whereas previously we had it coming from the AWS Metadata API, we'd now need it to be statically passed in to the method, is that ok?

I'm... not sure. I _think_ that, for now, the best way to handle this is actually to just have clients pass in the endpoint and then infer the region from the endpoint, with the endpoint defaulting to `https://sts.amazonaws.com` (and using the `us-east-1` region in that case). It's not elegant, but it makes fewer assumptions about AWS and thus should be less likely to be buggy.

As to why I recommend this:

First, sorry this is such a long explanation, but this is a pretty complex topic that's not well documented.

STS is [one of the weirdest AWS services](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html) -- it is "global" and has a single default global endpoint. But, unlike the other global services that I can think of off the top of my head (i.e., IAM and Route53), it has region-specific endpoints as well! AWS doesn't have great documentation about how to handle their home-spun authentication method in this case, so I did a bit of experimenting using boto3 (the python SDK, even though this is the Ruby SDK and Vault uses the golang SDK, and these are all corner cases that might not be well defined across SDKs) as it's the one I'm most familiar with.

Also an important note: the STS and IAM services are "global" but only within a "partition" and there are currently three partitions -- AWS Standard, GovCloud, and AWS China. It's unfortunately a bit tricky for users to test things out in one of the other partitions. Vault has some [awkward code](https://github.com/hashicorp/vault/blob/e9bbdd9c2c5aea2605c06cbf52be41aacc80d12b/builtin/credential/aws/backend.go#L211-L222) to try to handle this.

In this context, the region is needed for two reasons:

1. It's embedded in the 'credential scope" that is used in the authentication process (e.g., there's a string somewhere in the signing process that looks like `20180106/us-east-1/sts/aws4_request` that is included in the authorization header and is covered by the signature)
1. It can be used to determine which endpoint to use; further, the endpoint actually gets embedded in the request via the `Host` header which, as I mentioned above, is one of the required headers.

The second point is a bit more subtle. The actual endpoint that the Vault server uses is determined solely by the Vault server and not at all by the client, [defaulting](https://github.com/hashicorp/vault/blob/e9bbdd9c2c5aea2605c06cbf52be41aacc80d12b/builtin/credential/aws/path_login.go#L1130) to `https://sts.amazonaws.com` This is to prevent the Vault server from being abused by clients as a form of an open proxy. It looks like, when using the default endpoint of `https://sts.amazonaws.com` AWS expects a region in the credential scope of `us-east-1`.

So, if Vault admins do nothing, my proposed solution (defaulting to the endpoint of `https://sts.amazonaws.com` and defaulting to the region of `us-east-1`) **should** just work (I know, famous last words). And if the Vault server is configured to use a different STS endpoint, the clients need to know it to generate and sign the requests appropriately, and passing in the endpoint is how they would do it.

lib/vault/api/auth.rb
```diff
@@ -186,6 +190,59 @@ def aws_ec2(role, pkcs7, nonce = nil)
...
+      require "base64"
+
+      request_body   = 'Action=GetCallerIdentity&Version=2011-06-15'
+      request_url    = 'https://sts.amazonaws.com/'
+      request_method = 'POST'
+
+      vault_headers = {
+        'User-Agent' => Vault::Client::USER_AGENT,
+        'Content-Type' => 'application/x-www-form-urlencoded; charset=utf-8',
+        'X-Vault-AWSIAM-Server-Id' => iam_auth_header_value
```
joelthompson: It's actually [X-Vault-AWS-IAM-Server-ID](https://github.com/hashicorp/vault/blob/e9bbdd9c2c5aea2605c06cbf52be41aacc80d12b/builtin/credential/aws/path_login.go#L1620) (hyphen between AWS and IAM)

maschwenk:
@joelthompson Let me know if above commits follow your guidance above. Thanks so much for the thorough explanation. Maybe it makes sense to link to your PR/comment with regards to the STS endpoint.

lib/vault/api/auth.rb
```diff
@@ -186,6 +190,63 @@ def aws_ec2(role, pkcs7, nonce = nil)
...
+    # @param [String] sts_endpoint optional
+    #   https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
+    # @param [String] iam_auth_header_value optional
+    #
+    # @return [Secret]
+    def aws_iam(role, credentials_provider, sts_endpoint = 'https://sts.amazonaws.com', iam_auth_header_value = IAM_SERVER_ID_HEADER)
+      require "aws-sigv4"
+      require "base64"
+
+      valid_sts_endpoint = %r{https:\/\/sts.?(.*).amazonaws.com}.match(sts_endpoint)
```
joelthompson: FWIW, the STS endpoint for STS in the [China (Beijing) region (cn-north-1)](http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#cnnorth_region) is `sts.cn-north-1.amazonaws.com.cn`. This will still work because your match doesn't have a `$` at the end, but might be worth commenting so that future people don't add it and inadvertently break support for cn-north-1.

The [endpoint for GovCloud](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html) is `sts.us-gov-west-1.amazonaws.com` so it should work exactly as intended.

lib/vault/api/auth.rb
```diff
@@ -13,6 +13,10 @@ def auth
   end
 
   class Authenticate < Request
+
+    # canary header used for aws_ec2_iam
+    IAM_SERVER_ID_HEADER = "canaryHeaderValue".freeze
```
joelthompson: I'm not a fan of this. Right now, it happens to work in Vault that if you haven't configured the server ID header value, it will accept any or no header value. This isn't great behavior, and might be something that I'd want to change in the future, so I'd rather not have the ruby client depend on it. Can you just make the default value `nil` and then just add the header value into the `vault_headers` if it's non-`nil`.

lib/vault/api/auth.rb
```diff
@@ -186,6 +190,63 @@ def aws_ec2(role, pkcs7, nonce = nil)
...
+        url: sts_endpoint,
+        headers: vault_headers,
+        body: request_body
+      ).headers
+
+      payload = {
+        role: role,
+        iam_http_request_method: request_method,
+        iam_request_url: Base64.strict_encode64(sts_url),
+        iam_request_headers: Base64.strict_encode64(vault_headers.merge(sig4_headers).to_json),
```
joelthompson: This is probably OK, but just want to call out that this format for encoding the headers is only compatible with Vault starting with v0.8.0.

spec/integration/api/auth_spec.rb
```diff
@@ -211,5 +212,39 @@ module Vault
...
+      after do
+        subject.token = old_token
+      end
+
+      it "authenticates and saves the token on the client" do
+        expect(subject).to receive(:post).and_return 'huzzah!'
+        expect(Secret).to receive(:decode).and_return secret
+        expect(::Aws::Sigv4::Signer).to(
+          receive(:new).with(
+            service: 'sts', region: 'cn-north-1', credentials_provider: credentials_provider
```
maschwenk: I think this expectation could obviate the need for the comment in e598ffc

lib/vault/api/auth.rb
```diff
@@ -186,6 +186,66 @@ def aws_ec2(role, pkcs7, nonce = nil)
...
+    # @param [String] sts_endpoint optional
+    #   https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
+    # @return [Secret]
+    def aws_iam(role, credentials_provider, iam_auth_header_value = nil, sts_endpoint = 'https://sts.amazonaws.com')
+      require "aws-sigv4"
+      require "base64"
+
+      # STS in the China (Beijing) region (cn-north-1) is sts.cn-north-1.amazonaws.com.cn
+      # Take care changing below regex with that edge case in mind
+      valid_sts_endpoint = %r{https:\/\/sts.?(.*).amazonaws.com}.match(sts_endpoint)
```
joelthompson: I think you need to escape the . to be correct: `https:\/\/sts\.?(.*)\.amazonaws\.com`

spec/integration/api/auth_spec.rb
```diff
@@ -211,5 +212,39 @@ module Vault
         }.to_not change { subject.token }
       end
     end
+
+    describe "#aws_iam" do
+      before(:context) do
+        vault_test_client.sys.enable_auth("aws", "aws", nil)
```
joelthompson: I think it would be good to also configure the header value here and then insert it, just to ensure it gets processed properly.

lib/vault/api/auth.rb
```diff
@@ -186,6 +186,66 @@ def aws_ec2(role, pkcs7, nonce = nil)
...
+    # for future requests.
+    #
+    # @example
+    #   Vault.auth.aws_ecs_iam("dev-role-iam", Aws::AssumeRoleCredentials.new, "vault.example.com", "https://sts.us-east-2.amazonaws.com") #=> #<Vault::Secret lease_id="">
+    #
+    # @param [String] role
+    # @param [CredentialProvider] credentials_provider
+    #   https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CredentialProvider.html
+    # @param [String] iam_auth_header_value optional
+    #   As of Jan 2018, Vault will accept ANY or NO header, but this is subject to change and should not be relied upon
```
joelthompson: Would be better to read: "Vault will accept ANY or NO header if none is configured by the Vault server admin"

maschwenk:
@joelthompson Do you have anyone who would want to look at this in its updated state?

spec/integration/api/auth_spec.rb
```diff
@@ -216,6 +216,7 @@ module Vault
     describe "#aws_iam" do
       before(:context) do
         vault_test_client.sys.enable_auth("aws", "aws", nil)
+        vault_test_client.sys.put_auth_tune("aws", "iam_server_id_header_value" => "iam_header_canary")
```
joelthompson: I think the `.sys.put_auth_tune` here isn't right. The configuration of the header value is done via the [`auth/aws/config/client`](https://www.vaultproject.io/api/auth/aws/index.html#iam_server_id_header_value) API endpoint, not via a mount-tune call.

lib/vault/api/auth.rb
```diff
@@ -275,5 +269,21 @@ def tls(pem = nil, path = 'cert')
...
     end
+
+    private
+
+    # Parse an AWS region from a STS endpoint
+    # STS in the China (Beijing) region (cn-north-1) is sts.cn-north-1.amazonaws.com.cn
+    # Take care changing below regex with that edge case in mind
+    #
+    # @param [String] sts_endpoint
+    #   The raw pem contents to use for the login procedure.
```
joelthompson: This looks like something left over from copy/pasting?

lib/vault/api/auth.rb
```diff
@@ -186,6 +186,60 @@ def aws_ec2(role, pkcs7, nonce = nil)
       return secret
     end
 
+    # Authenticate via IAM EC2 method by providing a AWS CredentialProvider (either ECS, AssumeRole, etc.)
```
joelthompson: It would be better to say something like "Authenticate via AWS IAM auth method" as this doesn't use the ec2 auth method at all.

sethvargo:
I'm no longer employed at HashiCorp, so I can't help you. Sorry

flyinbutrs:
Oops, thanks for responding anyway though!

maschwenk:
@flyinbutrs @sethvargo I need to respond to Joel's good feedback sometime this weekend, then hopefully a Vault employee can review it

maschwenk:
@flyinbutrs I've responded to Joel's feedback. I also need to do a little manual integration testing myself, in my current project I used the initial implementation I had here but have not swapped out for this final solution

maschwenk:
Just did some integration testing on our own projects and this works excellently (currently just monkeypatching it in). I'd be great to get some feedback from a current employee of Hashicorp /cc @evanphx

maschwenk:
@flyinbutrs Apologies, I'm hesitant to make any fixes until I get some indication that this might get merged....

flyinbutrs:
No worries, understood. I'm using it off of your branch, and it's working great! Hopefully someone from hashicorp can take a look soon.

@briancain? I see you active from Hashicorp on other Ruby repos...  Any chance you can poke the right person for vault-ruby? Or maybe @jefferai since you seem to be a maintainer on vault itself?

maschwenk:
@evanphx I saw you merged a bunch of PR's on this repo this week. Any way you could cut a new release? I have a dependency on this in a gemspec so I don't have the option of pointing to master. Thanks!

PR #162, opened 2018-01-04
Rubocop

maschwenk:
In working on #161 I've noticed this repo's style is quite specific. I was wondering if you guys would consider standardizing the style you guys use through a Rubocop config? 

This initial commit just makes it so that all Rubocop rules pass, I wanted to put this here as a PR before I actually went ahead and actually made some suggestions on corrections.

Basically, I'd just look at the cases where there are just 1 or 2 occurrences of the violation and fixing them. If there are instances where you guys seem to be making an intentional choice of personal style, for instance:

```yml
# Offense count: 52
# Cop supports --auto-correct.
# Configuration parameters: AllowMultipleReturnValues.
Style/RedundantReturn:
  Enabled: false
```
e.g.
```ruby
    def app_id(app_id, user_id, options = {})
      .
      .
      return secret
    end
```

That'd be something I'd leave alone.

Let me know your thoughts, totally understand if it's not to your preference. I find having a Rubocop config makes it easier for outside contributors to not spend a bunch of time wondering if their PR will fit into your style guide.


evanphx:
Rubocop has so many things that I disagree with I don't really want to constantly maintain an exclusions list. If we were to use it, I'd want a global exclusions list that applies to all files, no having to set them per file.

maschwenk:
Understood

PR #167, opened 2018-02-28
Check Server health

evralston:
This was a byproduct of trying to shake out compatibility issues with newer Vault servers and the existing library. Not sure if it'd be helpful to upstream or not.

It's my first non-Chef Ruby so I am absolutely sure I've broken style guides for the code and the tests. Ready and willing to make changes or just toss it.


evralston:
(mostly wanted to see travis results of the compatibility matrix, since 0.5.x won't run on current MacOS)

PR #177, opened 2018-04-27
Add kubernetes auth method

ls-brentsmith:
Add functionality describe here https://www.vaultproject.io/docs/auth/kubernetes.html to authenticate with kubernetes services accounts.


evanphx:
Could you add some tests for this?

ls-brentsmith:
@evanphx sorry, I don't have the bandwidth.

CharlieC3:
@ls-brentsmith @evanphx Any chance we can get an update on this? Would be great to see this merged.

carlwiedemann:
Seems like this could be closed as a dupe of https://github.com/hashicorp/vault-ruby/pull/202

Thoughts?

PR #180, opened 2018-07-26
AWS IAM Route Support

computeracer:
Adds option to put in custom route to support different mount points.


universam1:
Great job, this is really missing. 
+1 to get that in

mflococodev:
Would be really useful. Thanks for adding

computeracer:
@evanphx Saw you did some of the PRs in the past. Would you be able to do this one?

Thanks,

Dan

computeracer:
@jefferai or @evanphx Any thoughts on this one?

Thank you for your time.

PR #181, opened 2018-07-30
support HTTP basic authn

nferch:
I expose Vault on a semi-untrusted network using a reverse proxy with HTTP authentication. The Vault gem didn't honor an address with a username and password in the URL.


nferch:
Thanks!

PR #183, opened 2018-09-05
Added support for namespaces during client initialization

stenio123:
Update to support new Namespaces feature of Vault 0.11.0.

Full sys/namespace support not included.

### DIFF
--- a/lib/vault/defaults.rb
+++ b/lib/vault/defaults.rb
@@ -11,6 +11,10 @@ module Vault
     # @return [String]
     VAULT_DISK_TOKEN = Pathname.new("#{ENV["HOME"]}/.vault-token").expand_path.freeze
 
+    # Namespace from env variable
+    # @return [String]
+    VAULT_NAMESPACE = ENV["VAULT_NAMESPACE"]
+
     # The list of SSL ciphers to allow. You should not change this value unless
     # you absolutely know what you are doing!
     # @return [String]
@@ -59,6 +63,13 @@ module Vault
         end
 
         nil
+      end
+
+
+      # Vault Namespacem if any.
+      # @return [String, nil]
+      def namespace
+        ENV["VAULT_NAMESPACE"]
       end
 
       # The SNI host to use when connecting to Vault via TLS.
--- a/spec/unit/defaults_spec.rb
+++ b/spec/unit/defaults_spec.rb
@@ -57,6 +57,20 @@ module Vault
       end
     end
 
+    describe ".namespace" do
+      it "uses ENV['VAULT_NAMESPACE'] if present" do
+        with_stubbed_env("VAULT_NAMESPACE" => "test") do
+          expect(Defaults.namespace).to eq("test")
+        end
+      end
+
+      it "falls back to the default VAULT_NAMESPACE" do
+        with_stubbed_env("VAULT_ADDR" => nil) do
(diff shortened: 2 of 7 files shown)


stenio123:
Example on how to validate can be found here:
https://gist.github.com/stenio123/30c6a444576b7d4401c473a9ed4aeaa9

chrishoffman:
One question I have here is should this be a parameter of the client or should you be able to pass it per request.  Both have their advantages but for some workflows, creating a new client seems like a heavy way for a API client to act.

lib/vault/defaults.rb
```diff
@@ -61,6 +65,13 @@ def token
         nil
       end
 
+
+      # Vault Namespacem if any.
```
kalafut: s/Namespacem/Namespace

spec/unit/defaults_spec.rb
```diff
@@ -57,6 +57,20 @@ module Vault
...
 
+    describe ".namespace" do
+      it "uses ENV['VAULT_NAMESPACE'] if present" do
+        with_stubbed_env("VAULT_NAMESPACE" => "test") do
+          expect(Defaults.namespace).to eq("test")
+        end
+      end
+
+      it "falls back to the default VAULT_NAMESPACE" do
+        with_stubbed_env("VAULT_ADDR" => nil) do
```
kalafut: Should be "VAULT_NAMESPACE", right?

lib/vault/defaults.rb
```diff
@@ -11,6 +11,10 @@ module Defaults
     # @return [String]
     VAULT_DISK_TOKEN = Pathname.new("#{ENV["HOME"]}/.vault-token").expand_path.freeze
 
+    # Namespace from env variable
+    # @return [String]
+    VAULT_NAMESPACE = ENV["VAULT_NAMESPACE"]
```
kalafut: Is this needed?

spec/unit/defaults_spec.rb
```diff
@@ -57,6 +57,20 @@ module Vault
       end
     end
 
+    describe ".namespace" do
```
kalafut: Isn't the desired behavior to use the env var if present, and to be nil if not, such that the header isn't added at all?

pablomolnar:
I had to monkey patch the lack of namespace support with the following code:

```ruby
DEFAULT_HEADERS = Vault::Client.class_eval('DEFAULT_HEADERS').dup.merge({"X-Vault-Namespace" => "my-namespace"})
Vault::Client.const_set('DEFAULT_HEADERS', DEFAULT_HEADERS)
```

Hoping namespace feature gets added soon...

mchugh19:
Hey all. Is there any hope of getting namespace support into the ruby lib? Once this is released, we'd still need to adjust our chef resources in order to take advantage, so this is a blocking issue.

eheydrick:
Would love to see this merged in. Looks like it needs a rebase at least.

isukapalli:
Is there any timeline when this gets merged and add support to Namespace? Thanks

Valarissa:
Hey all~! I'm new to HashiCorp, but for my first real assignment, I'm giving the Ruby Client some much needed TLC. I'll be starting by looking to merge this with some additional tests and other support for Namespaces like the ability to create Clients with different namespaces!

Stay tuned!

Valarissa:
Superseded by #229. All of @stenio123's changes are there and then some!

Issue #179, opened 2018-06-29
Can't connect to a TLS-enabled Vault on JRuby


dmaze:
We're using JRuby-9.1.16.0 against a TLS-enabled Vault.  The very most minimal Vault-client script:

```ruby
require 'vault'

Vault::Client.new.help('/')
```

Fails when you run it:

```shell-session
$ rvm use jruby-9.1.16.0@vault --create
$ gem install vault
$ VAULT_ADDR=https://localhost:8200 ruby tmp.rb
OpenSSL::SSL::SSLError: no cipher match
        ciphers= at org/jruby/ext/openssl/SSLContext.java:508
      set_params at org/jruby/ext/openssl/SSLContext.java:562
         connect at /Users/dmaze/.rvm/rubies/jruby-9.1.16.0/lib/ruby/stdlib/net/http.rb:903
        do_start at /Users/dmaze/.rvm/rubies/jruby-9.1.16.0/lib/ruby/stdlib/net/http.rb:868
           start at /Users/dmaze/.rvm/rubies/jruby-9.1.16.0/lib/ruby/stdlib/net/http.rb:863
           start at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/persistent.rb:698
  connection_for at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/persistent.rb:625
         request at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/persistent.rb:933
         request at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/client.rb:273
             get at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/client.rb:182
            help at /Users/dmaze/.rvm/gems/jruby-9.1.16.0@vault/gems/vault-0.11.0/lib/vault/api/help.rb:29
          <main> at tmp.rb:3
```

(Very low-level debugging with Wireshark indicates that this never gets past the initial TCP connection: _something_ must be listening on the indicated port but it doesn't necessarily need to be Vault or anything knowledgable about TLS; you get the same behavior if you launch a parallel `nc -l 8200` and point the script at that.)

[...]

PR #184, opened 2018-09-10
support sane defaults for jruby ssl_ciphers

mschuchard:
Updates the SSL_CIPHERS frozen string with logic for sane defaults for JRuby.

This is a quick fix for everyone running into #179 and who are unable to find the issue for assistance. A more complete fix would be appreciated afterward.

### DIFF
--- a/lib/vault/defaults.rb
+++ b/lib/vault/defaults.rb
@@ -14,7 +14,7 @@ module Vault
     # The list of SSL ciphers to allow. You should not change this value unless
     # you absolutely know what you are doing!
     # @return [String]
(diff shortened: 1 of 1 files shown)


mschuchard:
Soft ping. Puppet did work around this issue by outsourcing the function invocation to the client's MRI interpreter, but this would still be nice for people running into this issue while using JRuby for some other reason, or not running latest and great Puppet.

chrishoffman:
I believe it makes sense to keep this secure by default regardless of platform, especially as the security profiles of these underly platforms improve.

PR #188, opened 2018-11-14
Support for Vault 0.9-0.11, including versioned kv secrets.

pat:
This code could indeed be improved, and I'm pondering whether there's a way to have a single logical interface that handles both approaches, but here's what I've got for the moment:

```ruby
# Where /secret is a v1 KV secrets store (the existing behaviour):
client.logical(:unversioned).write("secret/bacon")
# Calling `logical` without an approach currently goes with the legacy (unversioned):
client.logical.write("secret/bacon")

# Where /secret is a v2 KV secrets store:
# The first argument gets split into two:
#   the mounted store path
#   the path of the secret in question
client.logical(:versioned).write("secret", "bacon")
# So, all operations via the versioned approach have this split-argument approach.
```

Tests pass on Vault v0.5 through to 0.11 🎉 

I feel like something smarter may mean introspection about the secret store versions, and that's perhaps not ideal? It's certainly extra overhead, albeit something that could be cached (but the extra smarts mean extra support challenges). My approach does allow for eventually making `:versioned` as the default when no argument is provided to `logical`, so at least that could be a viable deprecation/change moving forward in gem versions.

Would a retry-fallback be viable maybe? Attempt v2/versioned, and if that doesn't work, try v1/unversioned? Though `/data/` in-between the mount point and the secret names does make that messy (or even impossible?)

### DIFF
--- a/lib/vault/api/logical.rb
+++ b/lib/vault/api/logical.rb
@@ -1,231 +1,28 @@
-require_relative "secret"
-require_relative "../client"
-require_relative "../request"
-require_relative "../response"
+require_relative "logical/unversioned"
+require_relative "logical/versioned"
 
 module Vault
   class Client
+    LOGICAL_APPROACHES = {
+      unversioned: Logical::Unversioned,
+      versioned:   Logical::Versioned
+    }.freeze
+
     # A proxy to the {Logical} methods.
-    # @return [Logical]
-    def logical
-      @logical ||= Logical.new(self)
+    # @return [Logical::Unversioned, Logical::Versioned]
+    def logical(approach = :unversioned)
+      logicals[approach]
     end
 
-    def versioned_logical
-      @logical ||= VersionedLogical.new(self)
+    private
+
+    def logicals
+      @logicals ||= Hash.new do |hash, key|
+        hash[key] = LOGICAL_APPROACHES.fetch(key).new(self)
+      end
     end
   end
 
-  class Logical < Request
-    # List the secrets at the given path, if the path supports listing. If the
-    # the path does not exist, an exception will be raised.
-    #
-    # @example
-    #   Vault.logical.list("secret") #=> [#<Vault::Secret>, #<Vault::Secret>, ...]
-    #
-    # @param [String] path
-    #   the path to list
-    #
-    # @return [Array<String>]
-    def list(path, options = {})
-      headers = extract_headers!(options)
-      json = client.list("/v1/#{encode_path(path)}", {}, headers)
-      json[:data][:keys] || []
-    rescue HTTPError => e
-      return [] if e.code == 404
-      raise
-    end
-
-    # Read the secret at the given path. If the secret does not exist, +nil+
-    # will be returned.
-    #
-    # @example
-    #   Vault.logical.read("secret/password") #=> #<Vault::Secret lease_id="">
-    #
-    # @param [String] path
-    #   the path to read
-    #
-    # @return [Secret, nil]
-    def read(path, options = {})
-      headers = extract_headers!(options)
-      json = client.get("/v1/#{encode_path(path)}", {}, headers)
-      return Secret.decode(json)
-    rescue HTTPError => e
-      return nil if e.code == 404
-      raise
-    end
-
-    # Write the secret at the given path with the given data. Note that the
-    # data must be a {Hash}!
-    #
-    # @example
-    #   Vault.logical.write("secret/password", value: "secret") #=> #<Vault::Secret lease_id="">
-    #
-    # @param [String] path
-    #   the path to write
-    # @param [Hash] data
-    #   the data to write
-    #
-    # @return [Secret]
-    def write(path, data = {}, options = {})
-      headers = extract_headers!(options)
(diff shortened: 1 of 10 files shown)


lib/vault/api/logical.rb
```diff
@@ -1,150 +1,28 @@
...
-    # @param [String] path
-    #   the path to delete
-    #
-    # @return [true]
-    def delete(path)
-      client.delete("/v1/#{encode_path(path)}")
-      return true
+    # A proxy to the {Logical} methods.
+    # @return [Logical::Unversioned, Logical::Versioned]
+    def logical(approach = :unversioned)
```
grosser: might be better to require the argument to avoid confusion ?
also `logicals.fetch(approach)` to fail fast

pat: That'd be a breaking change, which I'm hesitant to do. At least this way it should be backwards compatible, and then future versions can mark an unprovided argument as deprecated, and then remove the old approach altogether.

lib/vault/api/logical/versioned.rb
```diff
@@ -0,0 +1,86 @@
...
+      rescue HTTPError => e
+        return [] if e.code == 404
+        raise
+      end
+
+      # Read the secret at the given path. If the secret does not exist, +nil+
+      # will be returned.
+      #
+      # @example
+      #   Vault.logical.read("secret/password") #=> #<Vault::Secret lease_id="">
```
ragurney: This documentation needs to be updated right? For example this should be something like `read("secret", "password")`, correct?

pat: Good catch, yes. If the general approach of this PR is fine and expected to be merged, then I'm very happy to add a commit updating the inline docs and the README.

ragurney: 👍 Thanks a lot for working on this, much appreciated!

lib/vault/api/logical/versioned.rb
```diff
@@ -0,0 +1,86 @@
...
+      #
+      # @param [String] path
+      #   the path to read
+      #
+      # @return [Secret, nil]
+      def read(mount, path, options = {})
+        headers = extract_headers!(options)
+        json = client.get("/v1/#{mount}/data/#{encode_path(path)}", {}, headers)
+        return Secret.decode(json[:data])
```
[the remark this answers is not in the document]
pat: Good idea! Just added this in.

lib/vault/api/logical/versioned.rb
pat: Just improved the inline docs a bit :)

grosser:
@evanphx  @chrishoffman can we please get this reviewed / merged ?

lib/vault/api/logical/versioned.rb
```diff
@@ -0,0 +1,94 @@
...
+      #   Vault.logical.delete("secret", "password") #=> true
+      #
+      # @param [String] mount
+      #   the mount point for the secret engine
+      # @param [String] path
+      #   the path to delete
+      #
+      # @return [true]
+      def delete(mount, path)
+        client.delete("/v1/#{mount}/data/#{encode_path(path)}")
```
ragurney: One more thought: as I'm playing around with this I'm wondering if it would make sense for this method to instead call v2's `destroy` instead of `delete` endpoint? It looks like in v2 this is more of a soft delete whereas v1's `delete` seems to be a hard delete, so making this call `destroy` might be more consistent 🤔

grosser:
v2 brought versioning, so if delete is a "versioned delete" then I think
that's what we want,
"real" delete (as in get rid of all versions) would be a new concept and
have it's own api call ?

chrishoffman:
Thanks for this contribution and sorry for the delay in review.  I think some changes need to be made here because this PR fully replaces the logical implementation with a versioned and unversioned logical.  Since versioning is only supported in KV, I think it may make sense to make a separate KV API that would transparently support v1 and v2 KV and leave logical unchanged.  This will also avoid some backwards compatibility issue that these changes introduce.

grosser:
I'd think transparent support is kinda hard (will take time to get right
and adds additional complexity), got any plan on how to do that ?

I don't see how this PR breaks anything since it still treats everything as
v1 unless told otherwise.

I'd love to get something merged even if it's a bit clunky to use at first
for v2 instead of us having to use a forked repo / write hacky code.

chrishoffman:
But it is not versioned logical, it is versioned kv (as a side note, we like to refer to this as v1 or v2 kv since versioning is just one part of the new api).  So I think a KV specific API makes more sense.  The Vault CLI already does this transparently using a preflight check in the `vault kv` command.  I am not suggesting that transparent support is required just that the integration point is incorrect.

grosser:
Ah ok :)
So "make a new kv class that supports v1 and v2" would be the correct PR ?

chrishoffman:
That's correct.  You can technically support kv v1 and v2 with the existing API.  Adding a kv class would be a nice addition.

pat:
Thanks for the feedback @chrishoffman - that makes sense. I'll see what I can do about a separate KV interface.

PR #196, opened 2018-12-19
Adds options for mounts

### LINKED ISSUE #195: Enhancement: Support `options` field on `/sys/mounts`
Right now, there's no way of performing certain logic based on if the `kv` endpoint is 1 or 2. 

I think this is just a case of adding 

```ruby
# @!attribute [r] options
    #   Options of the mount.
    #   @return [String]
    field :options
```

to the `lib/vault/api/sys/mount.rb` file

petems:
* Allows us to check what kv version is mounted
* Also allows us to retrieve options from existing 
mounts
* Restricts to 0.10+ for specs

Closes #195

### DIFF
--- a/lib/vault/api/sys/mount.rb
+++ b/lib/vault/api/sys/mount.rb
@@ -16,6 +16,11 @@ module Vault
     #   Type of the mount.
     #   @return [String]
     field :type
+
+    # @!attribute [r] options
+    #   Options of the mount.
+    #   @return [String]
+    field :options
   end
 
   class Sys < Request
@@ -44,6 +49,8 @@ module Vault
     #   the type of mount
     # @param [String] description
     #   a human-friendly description (optional)
+    # @param [Hash] options
+    #   mount type specific options that are passed to the backend (optional)
     def mount(path, type, description = nil, options = {})
       payload = options.merge type: type
       payload[:description] = description if !description.nil?
--- a/spec/integration/api/sys/mount_spec.rb
+++ b/spec/integration/api/sys/mount_spec.rb
@@ -12,6 +12,15 @@ module Vault
         expect(sys.type).to eq("system")
         expect(sys.description).to include("system endpoints")
       end
+
+      it "shows options for mounts" do
+        skip unless has_options_for_mount?
+        kv = subject.mounts[:secret]
+        expect(kv).to be_a(Mount)
+        expect(kv.type).to eq("kv")
+        expect(kv.description).to eq("key/value secret storage")
+        expect(kv.options).to eq(version: "2")
+      end
     end
 
     describe "#mount" do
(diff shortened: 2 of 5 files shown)


lib/vault/api/sys/mount.rb
```diff
@@ -44,8 +49,10 @@ def mounts
     #   the type of mount
     # @param [String] description
     #   a human-friendly description (optional)
-    def mount(path, type, description = nil)
-      payload = { type: type }
+    # @param [String] options
+    # mount type specific options that are passed to the backend (optional)
+    def mount(path, type, description = nil, options = {})
```
grosser: prefer that options includes `:description` ... maybe some fallback `if options is not a hash, then treat it as description`

lib/vault/api/sys/mount.rb
```diff
@@ -44,8 +49,10 @@ def mounts
     #   the type of mount
     # @param [String] description
     #   a human-friendly description (optional)
-    def mount(path, type, description = nil)
-      payload = { type: type }
+    # @param [String] options
```
grosser:
```suggestion
    # @param [Hash] options
```

lib/vault/api/sys/mount.rb
```diff
@@ -44,8 +49,10 @@ def mounts
...
+    # mount type specific options that are passed to the backend (optional)
```
grosser:
```suggestion
    #   mount type specific options that are passed to the backend (optional)
```

lib/vault/api/sys/mount.rb
```diff
@@ -44,8 +49,10 @@ def mounts
...
+    def mount(path, type, description = nil, options = {})
+      payload = options.merge({ type: type })
```
grosser:
```suggestion
      payload = options.merge(type: type)
```

spec/integration/api/sys/mount_spec.rb
```diff
@@ -12,6 +12,15 @@ module Vault
         expect(sys.type).to eq("system")
         expect(sys.description).to include("system endpoints")
       end
+
+      it "shows options for mounts" do
+        next unless has_options_for_mount?
```
grosser: could use skip so it's marked as skipped
```suggestion
        skip unless has_options_for_mount?
```

spec/integration/api/sys/mount_spec.rb
```diff
@@ -12,6 +12,15 @@ module Vault
...
+
+        kv = subject.mounts[:kv]
+        expect(kv).to be_a(Mount)
+        expect(kv.type).to eq("kv")
+        expect(kv.options).to eq(:version => 2)
```
grosser:
```suggestion
        expect(kv.options).to eq(version: 2)
```

spec/integration/api/sys/mount_spec.rb
```diff
@@ -22,6 +31,24 @@ module Vault
         expect(result.type).to eq("aws")
         expect(result.description).to eq("")
       end
+
+      it "returns the options from a mount" do
+        next unless has_options_for_mount?
+
+        expect(subject.mount("kv_v2_mount", "kv", 'kv v2 mount', options: {version: "2"})).to be(true)
```
grosser: isn't this supposed to be 
```suggestion
        expect(subject.mount("kv_v2_mount", "kv", 'kv v2 mount', version: "2")).to be(true)
```

petems: Nope, it's an options hash containing the entry of version . AFAIK :https://www.vaultproject.io/api/system/mounts.html#options

lib/vault/api/sys/mount.rb
petems: Not sure I understand this comment, do you mean that options contains a description? Or to remove the extra attribute?

grosser: both :D 

```ruby
def mount(path, type, options = {})
  description = (options.is_a?(Hash) ? options[:description] : options)
```

maxadamo:
will anyone push the merge button? :smiley_cat:

PR #200, opened 2019-04-04
Allow customisable auth path for all other auth methods

adammw:
Currently only tls and gcp auth methods support a custom path, and github, aws_ec2 and aws_iam support a custom endpoint. 

Adding path as a parameter to the remaining methods to allow them to be used with custom mount paths.


adammw:
cc @evanphx

Issue #197, opened 2019-01-28
Custom path in Auth.approle()


vaporz:
https://github.com/hashicorp/vault-ruby/blob/02f0532a802ba1a2a0d8703a4585dab76eb9d864/lib/vault/api/auth.rb#L95

Current path is fixed in code:"/v1/auth/approle/login".  
Can we support customizable path?  
For example, "/v1/auth/{MY_PATH}/login"

PR #206, opened 2019-08-15
Add Okta auth method

dylan-azucena:
Add functionality from here: https://www.vaultproject.io/docs/auth/okta.html to authenticate with Okta. Includes unit test (I believe integration tests would require an Okta account).

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -267,6 +267,28 @@ module Vault
       return secret
     end
 
+    # Authenticate via the okta authentication method. If authentication
+    # is successful, the resulting token will be stored on the client and used
+    # for future requests.
+    #
+    # @example
+    #   Vault.auth.okta("sethvargo", "s3kr3t") #=> #<Vault::Secret lease_id="">
+    #
+    # @param [String] username
+    # @param [String] password
+    # @param [Hash] options
+    #   additional options to pass to the authentication call, such as a custom
+    #   mount point
+    #
+    # @return [Secret]
+    def okta(username, password, options = {})
+      payload = { password: password }.merge(options)
+      json = client.post("/v1/auth/okta/login/#{encode_path(username)}", JSON.fast_generate(payload))
+      secret = Secret.decode(json)
+      client.token = secret.auth.client_token
+      return secret
+    end
+
     # Authenticate via a TLS authentication method. If authentication is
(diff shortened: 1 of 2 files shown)


robkinyon-roivant:
The problem with that PR is the same as problem #197 raises. You cannot assume `okta` is the correct path. We have multiple Okta backends, each with their own path. We already have to monkey-patch the LDAP authentication method to support non-standard paths.

dylan-azucena:
@robkinyon-roivant I believe what you're looking for is to pass the mount point via the custom options hash when using the auth method. Here's an example using the userpass auth method: https://github.com/hashicorp/vault-ruby/blob/e53f8ecbec06ca9456f14c449cbd5ea7da2eda0b/spec/integration/api/auth_spec.rb#L147
Have you tried this way of specifying the mount point with LDAP auth?

robkinyon-roivant:
> @robkinyon-roivant I believe what you're looking for is to pass the mount point via the custom options hash when using the auth method. Here's an example using the userpass auth method:
> 
> https://github.com/hashicorp/vault-ruby/blob/e53f8ecbec06ca9456f14c449cbd5ea7da2eda0b/spec/integration/api/auth_spec.rb#L147
> 
> 
> Have you tried this way of specifying the mount point with LDAP auth?

I have. The code doesn't look at `options[:mount]`. None of the code does, even the methods whose documentation says to use a custom mount-point, like `userpass()`.

https://github.com/hashicorp/vault-ruby/blob/e53f8ecbec06ca9456f14c449cbd5ea7da2eda0b/lib/vault/api/auth.rb#L142

Instead, that line (in `ldap()`) should be something like:
```ruby
      mount = options[:mount] || 'ldap'
      json = client.post("/v1/auth/#{mount}/login/#{encode_path(username)}", JSON.fast_generate(payload))
```
And then (with appropriate default) in `okta()`, `userpass()`, etc.

The mistake in the auth specs is that the `userpass` and `new-userpass` mounts both return the same thing. There's no way to differentiate between them, so you don't know that `options[:mount]` is being ignored. Instead, you should verify that `mount: not-a-thing` returns unauthenticated like on https://github.com/hashicorp/vault-ruby/blob/e53f8ecbec06ca9456f14c449cbd5ea7da2eda0b/spec/integration/api/auth_spec.rb#L151.

PR #210, opened 2019-09-25
Fix client#shutdown to not raise error when the pool is empty

fabioperrella:
## Motivation

The method `client#shutdown` is raising error when the pool is nil

## Proposal

```ruby
@nhp.shutdown() if @nhp
```

(I think the code above explains better than a description)


fabioperrella:
closing this to clean my list of open PRs

PR #223, opened 2020-04-02
Require VAULT_VERSION

mdeggies:
I noticed during testing that when `VAULT_VERSION` isn't set, the test suites run til completion and throws hard to read errors like: 

`{"request_id":"bf04957a-f3a4-3e9d-75b0-ee938aae17af","lease_id":"","renewable":false,"lease_duration":0,"data":null,"wrap_info":null,"warnings":["Invalid path for a versioned K/V secrets engine. See the API docs for the appropriate API endpoints to use. If using the Vault CLI, use 'vault kv put' for this operation."],"auth":null}`
 
It looks like this is required by all tests, so this change throws an abort when the var isn't set: 

Screen Shot 2020-04-02 at 4 51 30 PM

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -28,6 +28,7 @@ RSpec.configure do |config|
   # Disable real connections.
   config.before(:suite) do
     WebMock.disable_net_connect!(allow_localhost: true)
+    abort "Must set VAULT_VERSION environment variable" if (TEST_VAULT_VERSION.to_s == '')
   end
 
   # Ensure our configuration is reset on each run.


chrishoffman:
I'm not following on this one.  I think those error are legitimate since the behavior of the dev server in relation to the KV secret engine has changed with versions.  I believe there needs to be some work done to ensure you are using the correct version of the KV secret engine.

mdeggies:
Thanks for the fast review! Does `VAULT_VERSION` have to be set for any of the tests to pass? My understanding is that it does, otherwise they all fail with that same error:

```
  1) Vault::Logical#list returns all secrets
     Failure/Error: raise klass.new(address, response, [response.body])
     
     Vault::HTTPClientError:
       The Vault server at `http://127.0.0.1:8200' responded with a 404.
       Any additional information the server supplied is shown below:
     
         * {"request_id":"c1e47761-9eb8-b96a-c817-e16466371583","lease_id":"","renewable":false,"lease_duration":0,"data":null,"wrap_info":null,"warnings":["Invalid path for a versioned K/V secrets engine. See the API docs for the appropriate API endpoints to use. If using the Vault CLI, use 'vault kv put' for this operation."],"auth":null}
     
       Please refer to the documentation for help.
     # ./lib/vault/client.rb:405:in `error'
     # ./lib/vault/client.rb:291:in `request'
     # ./lib/vault/client.rb:203:in `put'
     # ./lib/vault/api/logical.rb:68:in `write'
     # ./spec/integration/api/logical_spec.rb:31:in `block (3 levels) in <module:Vault>'

  2) Vault::Logical#write allows special characters
     Failure/Error: raise klass.new(address, response, [response.body])
     
     Vault::HTTPClientError:
       The Vault server at `http://127.0.0.1:8200' responded with a 404.
       Any additional information the server supplied is shown below:
     
         * {"request_id":"cd5501f6-6485-d9a0-8d99-4a62c4989254","lease_id":"","renewable":false,"lease_duration":0,"data":null,"wrap_info":null,"warnings":["Invalid path for a versioned K/V secrets engine. See the API docs for the appropriate API endpoints to use. If using the Vault CLI, use 'vault kv put' for this operation."],"auth":null}
     
       Please refer to the documentation for help.
     # ./lib/vault/client.rb:405:in `error'
     # ./lib/vault/client.rb:291:in `request'
     # ./lib/vault/client.rb:203:in `put'
     # ./lib/vault/api/logical.rb:68:in `write'
     # ./spec/integration/api/logical_spec.rb:77:in `block (3 levels) in <module:Vault>'

...
```

So this is proposing we check this special var exists before running through the tests, otherwise they'd all fail and we wouldn't necessarily know the root cause. Let me know if I'm missing something, though. I'm not sure if this is the only case where `Invalid path for a versioned K/V secrets engine` is thrown, but in this case the server isn't running due to the vault version not being set. Maybe an alt option is just setting a default value here: https://github.com/hashicorp/vault-ruby/blob/master/spec/spec_helper.rb#L11

PR #225, opened 2020-04-13
Added vault version inference, updated webmock

Valarissa:
Version inference uses the response from 'vault --version' in order
to set what version is being tested against.

WebMock was updated to allow for newer versions of ruby.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -68,7 +68,7 @@ def versioned_kv_by_default?
 end
 
 def vault_is_enterprise?
-  vault_version_string.match(/\+i(ent|prem)$/)
+  vault_version_string.match(/\+(ent|prem)$/)
 end
 
 def with_stubbed_env(env = {})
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -64,6 +67,10 @@ def versioned_kv_by_default?
   Gem::Requirement.new(">= 0.10").satisfied_by?(TEST_VAULT_VERSION)
 end
 
+def vault_is_enterprise?
+  vault_version_string.match(/\+i(ent|prem)$/)
```
tvoran, line 71: Can you remind me what the `i` does here?

Valarissa: That... is probably a mistake from my vim using. I'll adjust it.

Valarissa: The idea is that it should match `+ent` and `+prem`, I probably accidentally added it when trying to type something.

PR #228, opened 2020-04-23
Locking aws-sigv4 at 1.1.1

Valarissa:
1.1.2 has changes that break the tests. As such, we're locking it here for now.


Valarissa:
@catsby `1.1.2` causes `vault-ruby`'s tests to break. We would need to update the tests appropriately.

PR #229, opened 2020-04-24
Adding support for namespaces

Valarissa:
This PR seeks to add support for namespaces.

Namespaces can be configured: 
 - A default can be provided as an environment variable on startup
 - Via an option when instantiating an instance of the `Vault::Client` class
 - Via the `Vault::Client#namespace=` method on an instance
 - As a passed in header per request to any given HTTP method of a `Vault::Client` instance
   - e.g. `client.put('/v1/secrets/sekkrit', {foo: "bar"}, {"X-Vault-Namespace": "baz"})`

In addition, the `/sys/namespaces` API has been implemented allowing you to create, update, read, delete, and list namespaces. This can be done like so:
 - `client.sys.create_namespace("foo")` POST
 - `client.sys.get_namespace("foo")` GET
 - `client.sys.namespaces` LIST
 - `client.sys.delete_namespace("foo")` DELETE


lib/vault/api/sys/namespace.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  class Sys
+    # List all namespaces in a given scope. Ignores nested namespaces.
+    #
+    # @example
+    #   Vault.sys.namespaces #=> { :foo => #<struct Vault::Namespace id="xxxx1", path="foo/" }
+    #
+    #   @return [Hash<Symbol, Namespace>]
+    #
+    #   NOTE: Due to a bug in Vault, to be fixed soon, this method CAN return a pure JSON string if a scoping namespace is provided.
```
catsby, line 22: It would be helpful if we had an issue number for this. Is it a bug in Vault OSS or Vault Ent? If OSS, let's make a bug and track it, and reference it here. If it's in Ent, then let's change this to "Due to a bug in Vault Enterprise". I think it's a good idea to reference (public) GH issues when mentioning bugs so when someone eventually comes around this code again, they can see if that bug is still relevant.

lib/vault/api/sys/namespace.rb
```diff
@@ -0,0 +1,81 @@
...
+    def namespaces(scoped=nil)
+      path = ["v1", scoped, "sys", "namespaces"].compact
+      json = client.list(path.join("/"))
+      json = json[:data] if json[:data]
+      if json[:key_info]
+        json = json[:key_info] if json[:key_info]
+        return Hash[*json.map do |k,v|
+          [k.to_s.chomp("/").to_sym, Namespace.decode(v)]
+        end.flatten]
```
catsby, line 31: This is a dense piece of code; When will the `if` on line 28 be false? When is `flatten` called here, is it called on the result of the lambda? And the result of that is then sent to `Hash`?

lib/vault/api/sys/namespace.rb
```diff
@@ -0,0 +1,81 @@
...
+    #
+    #   @return [Hash<Symbol, Namespace>]
+    #
+    #   NOTE: Due to a bug in Vault, to be fixed soon, this method CAN return a pure JSON string if a scoping namespace is provided.
+    def namespaces(scoped=nil)
+      path = ["v1", scoped, "sys", "namespaces"].compact
+      json = client.list(path.join("/"))
+      json = json[:data] if json[:data]
+      if json[:key_info]
+        json = json[:key_info] if json[:key_info]
```
catsby, line 28: If the `if` statement on line 27 is true, when would the `if` on line 28 be false?

lib/vault/api/sys/namespace.rb
```diff
@@ -0,0 +1,81 @@
...
+    #   Vault.sys.get_namespace("foo")
+    #
+    # @param [String] namespace
+    #   the path of the namespace ot be retrieved
+    #
+    # @return [Namespace]
+    def get_namespace(namespace)
+      json = client.get("/v1/sys/namespaces/#{namespace}")
+      json = json[:data] if json[:data]
+      json = json[:key_info] if json[:key_info]
```
catsby, line 77: Do these 2 statements append or overwrite each other? Maybe a code comment to clarify when we want `:data` and when we want (or would prefer?) `:key_info`

lib/vault/api/sys/namespace.rb
Valarissa: Forgot to clean that one up~! Whoops.

lib/vault/api/sys/namespace.rb
Valarissa: I pulled this directly from elsewhere in the codebase. I agree that it's super dense. I'll make it easier to read. :)

lib/vault/api/sys/namespace.rb
catsby: LGTM

lib/vault/api/sys/namespace.rb
catsby: 😄

lib/vault/api/sys/namespace.rb
catsby: I appreciate the effort of being consistent with other parts of the code! That said, the changes made in 89a18c358abf1abbf43ed657ee6b7eb94c418e6f are much easier to read and reason about 😄

PR #232, opened 2020-05-06
Relax aws-sigv4 dependency

lenon:
Hi! I noticed that vault-ruby 0.13.1 locked aws-sigv4 at 1.1.1. If I got it right, it was because of an issue with 1.1.2 that caused vault-ruby specs to fail. Since then, there was an update in aws-sigv4 1.1.3 that fixes the issue that was causing specs to fail:

https://github.com/aws/aws-sdk-ruby/pull/2291

In this PR I'm proposing to relax this dependency again and to remove the lock on 1.1.1. I also did a small change to replace regular doubles by verifying doubles so, in case of interface changes, there is a feedback about those changes.

### DIFF
--- a/spec/integration/api/auth_spec.rb
+++ b/spec/integration/api/auth_spec.rb
@@ -225,9 +225,14 @@ module Vault
 
       let!(:old_token) { subject.token }
       let(:credentials_provider) do
-        double(
-          credentials:
-            double(access_key_id: 'very', secret_access_key: 'secure', session_token: 'thing')
+        instance_double(
+          Aws::Sigv4::StaticCredentialsProvider,
+          credentials: instance_double(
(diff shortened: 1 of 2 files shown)


Valarissa:
I was literally about to push a change with 1.1.3 alone, but thank you SO much for updating the doubles to be instance doubles!

spec/integration/api/auth_spec.rb
```diff
@@ -225,9 +225,14 @@ module Vault
 
       let!(:old_token) { subject.token }
       let(:credentials_provider) do
-        double(
-          credentials:
-            double(access_key_id: 'very', secret_access_key: 'secure', session_token: 'thing')
+        instance_double(
```
Valarissa, line 228: This is so great. Thanks a lot.

PR #238, opened 2020-06-03
Add accessor to sys.auths response

Saraislet:
Hi! Returning the accessor with sys.auths would simplify some work we're doing for managing auth methods.

I added tests, with I think the reasonable assumption that the token auth mount is always created and that we don't need to test the value of the accessor since it is not feasibly predictable (I gather that Vault's golang tests also make this assumption).

The existing tests are somewhat bare. Is this sufficient, and reasonably in line with existing tests? Thanks : )


saraislet:
😑 I left the company since I signed the agreement, so the email address is no longer mine to use.

PR #242, opened 2020-07-07
Add support for resource quotas

Valarissa:
This PR adds resource quota definition and retrieval to the Vault gem.

It will now be possible to manage resource quotas per the HTTP API using calls in the gem directly.

## Usage

Creation/Update:
```
Vault.client.sys.create_quota("rate-limit", "INSERT_NAME_OF_CHOICE", rate: 167, burst: 300)
```

Deletion:
```
Vault.client.sys.delete_quota("rate-limit", "INSERT_NAME_OF_CHOICE")
```

Retrieval:
```
Vault.client.sys.get_quota("rate-limit", "INSERT_NAME_OF_CHOICE")
```

## Tests

Tests need to be run manually as of now. The reason being, this is a feature that will be released for 1.5. There is a filter in place that ignores the rate-limit tests with a vault version "< 1.5" and ignores the lease-count tests with an enterprise vault version of "< 1.5" This restriction can be circumvented by directly calling the tests via:
```
bundle exec rspec spec/integration/api/sys/quota_spec.rb:40
```
for rate-limits, and:
```
bundle exec rspec spec/integration/api/sys/quota_spec.rb:58
```
for lease-counts. Keep in mind, that the tests will fail unless you are using an appropriate version of Vault, like one built from source from the current branch.


lib/vault/api/sys/quota.rb
```diff
@@ -0,0 +1,75 @@
...
+      return true
+    end
+
+    def delete_quota(type, name)
+      path = generate_path(type, name)
+      client.delete(path)
+      return true
+    end
+
+    def get_quota(type, name)
```
calvn, line 43: Should we include a method for updating the config?

spec/integration/api/sys/quota_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "spec_helper"
+
+RSpec.shared_examples "quota specs" do |type|
```
calvn, line 3: Should we include a set of tests for creating/updating and reading the config?

lib/vault/api/sys/quota.rb
Valarissa: Done

spec/integration/api/sys/quota_spec.rb
Valarissa: Done

PR #243, opened 2020-07-30
Updates tests to support the released spec for resource quotas

Valarissa:
The resource quotas spec/behavior was changed before release, this change brings the tests in line with those changes.

### DIFF
--- a/spec/integration/api/sys/quota_spec.rb
+++ b/spec/integration/api/sys/quota_spec.rb
@@ -11,7 +11,6 @@ RSpec.shared_examples "quota specs" do |type|
 
   it "creates a #{type} quota" do
     subject.sys.create_quota(type, "test_1", create_args)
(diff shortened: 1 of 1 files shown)


spec/integration/api/sys/quota_spec.rb
```diff
@@ -11,6 +11,7 @@
 
   it "creates a #{type} quota" do
     subject.sys.create_quota(type, "test_1", create_args)
+    p subject.sys.get_quota(type, "test_1")
```
jasonodonnell, line 14: Not familiar with Ruby, wondering what the `p` is for?

Valarissa: It's to print things... which I forgot to remove... thanks.

Valarissa: Removed

PR #252, opened 2020-12-19
Make connection pool timeout configurable

sshock:
Make the connection pool timeout (default is 0.5 secs) configurable, similar to how the connection pool size is configurable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ Development
 Important Notes:
 
 - **All new features must include test coverage.** At a bare minimum, Unit tests are required. It is preferred if you include integration tests as well.
-- **The tests must be be idempotent.** The HTTP calls made during a test should be able to be run over and over.
+- **The tests must be idempotent.** The HTTP calls made during a test should be able to be run over and over.
 - **Tests are order independent.** The default RSpec configuration randomizes the test order, so this should not be a problem.
 - **Integration tests require Vault**  Vault must be available in the path for the integration tests to pass.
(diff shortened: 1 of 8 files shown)


sshock:
I'm also curious to know if we really want 0.5 to be the default connection pool (checkout) timeout.  That seems really fast.  Maybe 2 or even 5 seconds should be the default?  Let me know and I can tweak it if wanted.

Valarissa:
First, my apologies for not getting to this earlier. I had erroneously set up notifications and hadn't seen anything come in. To that end, mea culpa. I'll be taking a look at this in the very near future.

README.md
```diff
@@ -216,7 +216,7 @@ Development
 Important Notes:
 
 - **All new features must include test coverage.** At a bare minimum, Unit tests are required. It is preferred if you include integration tests as well.
-- **The tests must be be idempotent.** The HTTP calls made during a test should be able to be run over and over.
```
Valarissa, line 219: Thanks for this :D

lib/vault/persistent.rb
```diff
@@ -202,11 +202,6 @@ class PersistentHTTP
 
   HAVE_OPENSSL = defined? OpenSSL::SSL # :nodoc:
 
-  ##
-  # The default connection pool size is 1/4 the allowed open files.
-
-  DEFAULT_POOL_SIZE = 16
```
Valarissa, line 208: Great cleanup

Valarissa (approved):
Looks great. Thanks for putting the time into this. Apologies again about the delay in reviewing!

sshock:
Thanks!

PR #255, opened 2021-04-09
Fix DEFAULT_PARAMS adjustments based on OpenSSL support

Valarissa:
Previously, the OpenSSL DEFAULT_PARAMS were not being set correctly.
h/t to @nevans for the heads-up


austingebauer (approved):
LGTM!

Issue #254, opened 2021-03-30
vault-ruby substitute for /v1/auth/cert/login with a role


khawkins:
Per [the docs](https://www.vaultproject.io/docs/auth/cert) on TLS-based authentication, the `/v1/auth/cert/login` API endpoint accepts a role for the cert being used to authenticate. However, the [`Vault.auth.tls()` method](https://github.com/hashicorp/vault-ruby/blob/dac98febad779048bfa24c6452a946fd979754cf/lib/vault/api/auth.rb#L290) does not appear to support this input. And it seems to be required in our case as, if we authenticate with the `tls()` method, while we do successfully authenticate and get back an auth token, subsequent attempts to use it to authorize our requests fail with a 403. This doesn't happen if we use the token garnered from hitting the API endpoint directly (i.e. a raw HTTP request), specifying the proper role as part of process.

Could someone point out if I'm missing something obvious, or suggest another preferred workflow for using vault-ruby to get a token authorized for such a role? Thank you in advance!

khawkins:
Never mind. Turns out we were using the token wrong. I had the following code to set our token:
```ruby
 vault_token = Vault.auth.tls
 Vault.token = vault_token  # This is the problem
 secrets = Vault.kv('our_prefix').read('our_key_path')
```

`Vault.auth.tls` returns a `Vault::Secret`, while the token of `Vault::Client` is just a string, and we should be using `secret.auth.client_token` for that value instead. That's why we were getting 403s when using our token. I figured I'd document it here for posterity, in case someone else makes the same misstep.

khawkins:
Actually, while the above misstep is legitimate, the underlying issue is still there. It didn't come up for me in my test use case because I was testing against a vault with only one role, so it worked.

If you can't specify the role as part of authentication, you're at the mercy of returning credentials for whatever the default role is. If that's not the right role, you're out of luck.

This seems like a bug to me, unless there's another way to supply the role, after authentication. But it seems too late in the process to do so, to me.

Valarissa:
Hey there, just wanted to let you know I'm taking a look at this.

calvn:
Apologies for the delay on this! The PR has been merged and should be available in the next release.

PR #256, opened 2021-04-09
Add ability to specify a name for a cert to check for auth#tls

Valarissa:
This pull request seeks to add the ability to specify a certificate name when attempting TLS auth. This will bring the behavior more in line to the API specified here: https://www.vaultproject.io/api/auth/cert#login-with-tls-certificate-method

### DIFF
--- a/lib/vault/api/auth.rb
+++ b/lib/vault/api/auth.rb
@@ -287,11 +287,13 @@ module Vault
     #   The path to the auth backend to use for the login procedure.
     #
     # @return [Secret]
-    def tls(pem = nil, path = 'cert')
+    def tls(pem = nil, path = 'cert', name: nil)
       new_client = client.dup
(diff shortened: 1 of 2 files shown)


lib/vault/api/auth.rb
```diff
@@ -287,11 +287,13 @@ def gcp(role, jwt, path = 'gcp')
     #   The path to the auth backend to use for the login procedure.
     #
```
austingebauer, line 288: Might want to add an `@param` comment for the name parameter? Not sure if these comments are used to generate docs.

austingebauer (approved):
LGTM 👍

pcman312 (approved):

lib/vault/api/auth.rb
calvn: Added in d7586f0.

calvn:
Closes #254

PR #264, opened 2021-12-23
Adds KV PATCH support via 'update' methods.

phallstrom:
This adds an `update` method to `KV` to allow for patching paths as described here:
https://www.vaultproject.io/api-docs/secret/kv/kv-v2#patch-secret

It also adds an `update` method to `Logical` to allow for patching paths as well.  I did this for parity, but I'm not entirely sold on it as versioning requires the path start with `data` which the KV class does automatically, but Logical does not. I could easily be talked into removing that from this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,8 +71,8 @@ workflows:
               only: /^v[0-9]+\.[0-9]+\.[0-9]+.*/
           matrix:
             parameters:
-              ruby-version: ["2.7.1", "2.6", "2.5"]
-              vault-version: ["1.5.0", "1.4.2", "1.4.1", "1.4.0", "1.3.6"]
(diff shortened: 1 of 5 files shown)


phallstrom:
No idea why the Jira Sync is failing, but don't think that's due to my PR.

The test failures are legitimate, but maybe aren't important? I'm running 1.9.2 locally and they pass. I suspect the HTTP status codes have changed among vault versions. Happy to loosen that up, not sure how you'd like to handle it.

jdoss:
LGTM. Great job Phillip!

kalw:
Anything that we can do to help this one moving forward ?

phallstrom:
No one seems to care... I'm closing this just to stop from seeing it on my Github pages.

PR #266, opened 2022-04-26
[VAULT-5411] Add handling of 412 errors to with_retries

digivava:
Now that Vault 1.10 supports [server-side consistent tokens](https://www.vaultproject.io/docs/faq/ssct), clients need to be able to retry on the 412 error that is returned when the index state on the token does not match what's on the performance standby node.

I named the new error type similarly to how it is named in our Go library.

I also added this error type to the default list of retried exceptions, to be more aligned with the Go library, which automatically retries on 412s.

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -296,6 +296,8 @@ module Vault
           request(verb, response[LOCATION_HEADER], data, headers)
         when Net::HTTPSuccess
           success(response)
+        when Net::HTTPPreconditionFailed
+          raise MissingRequiredStateError.new
         else
           error(response)
         end
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -296,8 +296,6 @@ module Vault
           request(verb, response[LOCATION_HEADER], data, headers)
         when Net::HTTPSuccess
           success(response)
-        when Net::HTTPPreconditionFailed
-          raise MissingRequiredStateError.new
         else
           error(response)
         end
@@ -394,6 +392,8 @@ module Vault
 
       # Use the correct exception class
       case response
+      when Net::HTTPPreconditionFailed
+        raise MissingRequiredStateError.new
       when Net::HTTPClientError
         klass = HTTPClientError
       when Net::HTTPServerError
--- a/lib/vault/errors.rb
+++ b/lib/vault/errors.rb
@@ -22,6 +22,19 @@ EOH
     end
   end
 
+  class MissingRequiredStateError < VaultError
+    def initialize
+      super <<-EOH
+The performance standby node does not yet have the 
+most recent index state required to authenticate 
+the request.
+
+The request should be retried with the with_retries clause.
+
+EOH
+    end
+  end
+
   class HTTPConnectionError < VaultError
     attr_reader :address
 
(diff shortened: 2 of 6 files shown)


lib/vault/client.rb
```diff
@@ -296,6 +296,8 @@ def request(verb, path, data = {}, headers = {})
           request(verb, response[LOCATION_HEADER], data, headers)
         when Net::HTTPSuccess
           success(response)
+        when Net::HTTPPreconditionFailed
+          raise MissingRequiredStateError.new
```
chrisarcand, line 300: Does this belong in the `error()` method called in the `else` here, do you think? Seems like it as other `Net` exceptions are handled there as well. 

https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/client.rb#L381-L415

lib/vault/errors.rb
```diff
@@ -18,6 +18,19 @@ def initialize
...
+  end
+
+  class MissingRequiredStateError < VaultError
+    def initialize
+      super <<-EOH
+The performance standby node does not yet have the 
+most recent index state required to authenticate 
+the request.
+
+The request should be retried with the with_retries clause.
```
chrisarcand, line 32: * I've been quickly skimming the behavior of `RETRIED_EXCEPTIONS` and I _think_ that exception list is only retried with `with_retries`, is that right? An `HTTPServerError` is _not_ retried without the `with_retries` block, currently?

* If the above is true, I can see giving the actionable advice here at the end. Is this always true, though? Any reason a client may be wanting to see an explicit error returned here? Maybe `Generally, the request should be retried...`?

chrisarcand (changes requested):
Looks good, I just think the logic for this error may belong in a different spot, unless I'm missing something!

lib/vault/errors.rb
digivava: RETRIED_EXCEPTIONS seems to just be the default value used _if_ no error types were passed as parameters to `with_retries`. So if for some reason someone wanted to see the explicit error, they could either make a request without `with_retries`, or call `with_retries` with a list of params to retry on that doesn't include `MissingRequiredStateError`.

I agree though about the messaging, I'll change it to "Generally" like you said.

lib/vault/client.rb
```diff
@@ -392,6 +392,8 @@ def error(response)
 
       # Use the correct exception class
       case response
+      when Net::HTTPPreconditionFailed
+        raise MissingRequiredStateError.new
```
chrisarcand, line 396: If the 412 response from Vault actually has the context that you hardcoded in the message there “The performance standby node…“, generally speaking, I’d probably match the signatures and return the message from Vault directly.  That also avoids the slight weirdness of raising an exception in a method that sets everything to then raise an exception.

So you could probably catch the 412 and then let the existing logic work for you in HTTPError:

```
The Vault server at `#{address}' responded with a #{code}.
Any additional information the server supplied is shown below:
#{errors.join("\n").rstrip}
Please refer to the documentation for help.
```

Which I’m assuming would end up like this:

```
The Vault server at `#{address}' responded with a 412.
Any additional information the server supplied is shown below:

The performance standby node does not yet have the most recent....

Please refer to the documentation for help.
```

What does Vault actually return for this error? 

Otherwise, if Vault doesn’t actually return anything useful beyond the status code, what you have is probably fine.

digivava: Vault seems to not return any response body for this error, just the status code.

PR #268, opened 2022-05-17
Support IPv6 URIs

### LINKED ISSUE #248: Vault.address does not support IPv6 addresses
When setting Vault.address to 'http://[::1]:8200' I get the following error:

`Failed to open TCP connection to [::1]:8200 (getaddrinfo: Name or service not known)`

The vault agent is running happily:

```
# netstat -ntpl
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      928/sshd            
tcp6       0      0 ::1:8200                :::*                    LISTEN      2089/vault          
tcp6       0      0 :::22                   :::*                    LISTEN      928/sshd
```

wjordan:
Fixes #248 by changing `uri.host` to `uri.hostname` in `Vault::PersistentHTTP`.

The underlying issue is that [`URI#host`](https://ruby-doc.org/stdlib/libdoc/uri/rdoc/URI/Generic.html#host) is the incorrect method to use for passing a hostname to `Net::HTTP`, because it doesn't handle IPv6 addresses correctly:

> Since IPv6 addresses are wrapped with brackets in URIs, this method returns IPv6 addresses wrapped with brackets. This form is not appropriate to pass to socket methods such as `TCPSocket.open`. If unwrapped host names are required, use the [`hostname`](https://ruby-doc.org/stdlib/libdoc/uri/rdoc/URI/Generic.html#method-i-hostname) method.

Note: this module appears to be vendored from [`drbrain/net-http-persistent`](https://github.com/drbrain/net-http-persistent), which already included a fix for this issue (https://github.com/drbrain/net-http-persistent/pull/76, https://github.com/drbrain/net-http-persistent/pull/82, https://github.com/drbrain/net-http-persistent/commit/aa112b27d435e21c978f8274fe72d259db93aa2d) in its latest release.

### DIFF
--- a/lib/vault/persistent.rb
+++ b/lib/vault/persistent.rb
@@ -603,10 +603,10 @@ class PersistentHTTP
   def connection_for uri
     use_ssl = uri.scheme.downcase == 'https'
 
-    net_http_args = [uri.host, uri.port]
+    net_http_args = [uri.hostname, uri.port]
 
     net_http_args.concat @proxy_args if
-      @proxy_uri and not proxy_bypass? uri.host, uri.port
+      @proxy_uri and not proxy_bypass? uri.hostname, uri.port
 
     connection = @pool.checkout net_http_args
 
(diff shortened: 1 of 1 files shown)


jphenow (approved):

swenson (approved):
LGTM

PR #283, opened 2023-03-14
Added configuration option `Vault.allowed_proxy_addresses`

tbehling:
This PR adds a configuration option `Vault.allowed_proxy_addresses`, an array of hostnames, to allow `vault-ruby` to use a proxy with an HTTPS connection.

Currently, `vault-ruby` explicitly refuses to connect if a proxy is configured and the Vault target is HTTPS:

https://github.com/hashicorp/vault-ruby/blob/42d21087b21da335e4ef15469c18755acc875910/lib/vault/client.rb#L249-L251

I presume (this code is several years old) this was intended as a partial protection against a MITM attack.  However, sometimes it's necessary to transit a proxy to reach Vault.  When HTTPS is used, the connection is still end-to-end protected; an HTTPS proxy does not unwrap the TLS stream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v?.??.? (Unreleased)
 
+IMPROVEMENTS
+
+- Added configuration option `Vault.allowed_proxy_addresses` so a proxy can be specified even with an HTTPS connection.
+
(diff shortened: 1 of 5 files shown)


tbehling:
Closing this in favor of #284, which just removes the check instead of creating a new configuration option to skip it.

PR #284, opened 2023-03-21
Allow using an HTTPS connection with a proxy

tbehling:
This PR allows `vault-ruby` to use a proxy with an HTTPS connection to Vault.

Currently, vault-ruby explicitly refuses to connect if a proxy is configured and the Vault target is HTTPS.  I presume (this code is several years old) this check was intended as a partial protection against a MITM attack. However, sometimes it's necessary to transit a proxy to reach Vault.  When HTTPS is used, the connection is still end-to-end protected; an HTTPS proxy does not unwrap the TLS stream.

For comparison, I could find no such restriction on HTTP proxy usage in Vault's [Go client](https://github.com/hashicorp/vault/blob/main/api/client.go).

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -61,11 +61,6 @@ module Vault
       a << PersistentHTTP::Error
     end.freeze
 
-    # Indicates a requested operation is not possible due to security
-    # concerns.
-    class SecurityError < RuntimeError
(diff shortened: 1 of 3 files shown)


evanphx (approved):
You're the boss boss!

PR #294, opened 2023-06-26
Ci refactor

dhuckins:
The main change here is to not read the `~/.vault-token` file for the token, just hard-code it.
this takes out the race condition that was happening and allows us to have more consistent CI runs

### DIFF
--- a/spec/integration/api/sys/namespace_spec.rb
+++ b/spec/integration/api/sys/namespace_spec.rb
@@ -73,7 +73,7 @@ module Vault
     describe "unsupported namespaces", non_ent_vault: ">= 0.13" do
       context "#namespace" do
         it "returns a Vault::HTTPClientError with a note that the path is unsupported" do
-          expect{ subject.sys.get_namespace("foo") }.to raise_error(Vault::HTTPClientError, /unsupported path/)
+          expect{ subject.sys.get_namespace("foo") }.to raise_error(Vault::HTTPClientError)
         end
       end
     end
--- a/spec/support/vault_server.rb
+++ b/spec/support/vault_server.rb
@@ -42,6 +42,7 @@ module RSpec
         io.unlink
       end
       wait_for_ready
+      puts "vault server is ready"
 
       @token = File.read(TOKEN_PATH)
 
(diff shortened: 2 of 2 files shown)


spec/integration/api/sys/namespace_spec.rb
```diff
@@ -73,7 +73,7 @@ module Vault
     describe "unsupported namespaces", non_ent_vault: ">= 0.13" do
       context "#namespace" do
         it "returns a Vault::HTTPClientError with a note that the path is unsupported" do
-          expect{ subject.sys.get_namespace("foo") }.to raise_error(Vault::HTTPClientError, /unsupported path/)
+          expect{ subject.sys.get_namespace("foo") }.to raise_error(Vault::HTTPClientError)
```
dhuckins, line 76: we changed the text of the error at some point

spec/support/vault_server.rb
```diff
@@ -37,39 +41,45 @@ def initialize
...
     def address
       "http://127.0.0.1:8200"
     end
 
-    def wait_for_ready(&block)
-      Timeout.timeout(15) do
-        while !File.exist?(TOKEN_PATH)
-          sleep(0.25)
+    def wait_for_ready
+      uri = URI(address + "/v1/sys/health")
```
averche, line 69: This is much better 👍

spec/support/vault_server.rb
```diff
@@ -37,39 +41,45 @@ def initialize
...
+      Timeout.timeout(10) do
+        loop do
+          begin
+            response = Net::HTTP.get_response(uri)
+            return true if response.code != 200
```
averche, line 74: I don't really know ruby syntax but would this return false on non-200 response code or not return at all?

spec/support/vault_server.rb
```diff
@@ -28,7 +28,11 @@ def initialize
       end
 
       io = Tempfile.new("vault-server")
-      pid = Process.spawn({}, "vault server -dev", out: io.to_i, err: io.to_i)
+      pid = Process.spawn(
+        { "VAULT_DEV_ROOT_TOKEN_ID" => "root" },
+        "vault server -dev",
```
averche, line 33: might be easier to pass it as command-line flag:

```suggestion
        "vault server -dev -dev-root-token-id=root",
```

averche (approved):

AnPucel (approved):

spec/support/vault_server.rb
dhuckins: would not return at all 
perhaps 
```suggestion
            if response.code != 200
              return true
            end
```

would be better?

averche: Yes, definitely more clear for a ruby novice like me 😄

PR #297, opened 2023-09-13
TF-8697: Support TLS 1.3

### LINKED ISSUE #263: Failure to connect with TLS v1.3
Hi! I have set up a server with only TLSv1.3 support and I can't connect with the Ruby client.

The problem is that in the following line the TLS version is hardcoded as 1.2

https://github.com/hashicorp/vault-ruby/blob/714a30516ac9d5c5477a8de549464d5a8aabd9b9/lib/vault/client.rb#L117

The Ruby OpenSSL maintainers say [this shouldn't be used](https://github.com/ruby/openssl/issues/473#issuecomment-964252959), instead we should just set a minimum version using the [min_version method](https://ruby-doc.org/stdlib-3.0.3/libdoc/net/http/rdoc/Net/HTTP.html#min_version).

If I change `connection.ssl_version` to `connection.min_version`:

https://github.com/hashicorp/vault-ruby/blob/714a30516ac9d5c5477a8de549464d5a8aabd9b9/lib/vault/persistent.rb#L1047

and `"TLSv1_2"` to `:TLS1_2` here:

https://github.com/hashicorp/vault-ruby/blob/714a30516ac9d5c5477a8de549464d5a8aabd9b9/lib/vault/client.rb#L117

the problem gets fixed and I can connect to the server :-)

Of course, you'll probably want to change `ssl_version` to `min_ssl_version` or something like that to reflect the actual semantics of the constraint. Or, if you really want to constraint to a specific SSL version then you should set both, `min_version` and `max_version`, and I guess it would be a good idea to let the user of the library choose the constraint.

sudomateo:
Closes #263

This pull request moves away from using the [deprecated `ssl_version`](https://docs.ruby-lang.org/en/master/OpenSSL/SSL/SSLContext.html#method-i-ssl_version-3D) to using [`min_version`](https://docs.ruby-lang.org/en/master/OpenSSL/SSL/SSLContext.html#method-i-min_version-3D) in order to support TLS 1.3. Previously, the code statically defined the TLS version to TLS 1.2, meaning that this Vault client could not connect to Vault servers that only spoke TLS 1.3.

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -112,8 +112,8 @@ module Vault
 
         @nhp.verify_mode = OpenSSL::SSL::VERIFY_PEER
 
-        # Vault requires TLS1.2
-        @nhp.ssl_version = "TLSv1_2"
+        # Vault requires at least TLS1.2
+        @nhp.min_version = OpenSSL::SSL::TLS1_2_VERSION
 
         # Only use secure ciphers
         @nhp.ciphers = ssl_ciphers
--- a/lib/vault/persistent.rb
+++ b/lib/vault/persistent.rb
@@ -74,11 +74,11 @@ autoload :OpenSSL, 'openssl'
 # #ca_path            :: Directory with certificate-authorities
 # #cert_store         :: An SSL certificate store
 # #ciphers            :: List of SSl ciphers allowed
+# #min_version        :: Minimum SSL version to use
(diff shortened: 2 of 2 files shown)


averche (approved):
This makes sense to me though I don't have much experience with ruby

evanphx (approved):
ship

Issue #313, opened 2023-11-23
Vault gem not installing in Ruby 2.3.1 due to aws-sigv4 dependencies


kumar1202:
The vault gem uses the aws-sigv4 gem in the [gemspec](https://github.com/hashicorp/vault-ruby/blob/master/vault.gemspec#L24) and has not been locked.

Yesterday, the vault gem started failing in Ruby 2.3 when the aws-eventstream and aws-sigv4 gems were upgraded to [1.3.0](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-eventstream/CHANGELOG.md#130-2023-11-22) and [1.7.0](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sigv4/CHANGELOG.md#170-2023-11-22) respectively.

Although, the README says that Vault is still is supported in Ruby versions >2.0.0. The release of aws gems used as dependencies are breaking the vault in EOL Ruby environments.

Locking the aws-sigv4 version to the earliest available versions also won't help because the aws-eventstream lock was introduced as `'~> 1.0', '>= 1.0.2')` [here](https://github.com/aws/aws-sdk-ruby/commit/1f2aa49f91787fdd0597911cfc8e759d72daa712#diff-b5eed504a8582c5f562a40fc9faf9041b66cd9346694cb911ec5f61511a130dfR11) which still points to the latest version and eventually breaks on Ruby 2.3.

The solution is to lock, aws-sigv4 to 1.6.0 and aws-eventstream to 1.2.0 for Ruby versions < 2.4.0 in the vault.gemspec.

kumar1202:
We are still not able to install the gem on Ruby 2.3 when the source is RubyGems, as it didn't create the dependencies correctly. -> https://rubygems.org/gems/vault/versions/0.18.2

```
~# /opt/chef/embedded/bin/ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]

~# /opt/chef/embedded/bin/gem install vault
Fetching: aws-eventstream-1.3.0.gem (100%)
ERROR:  Error installing vault:
    aws-eventstream requires Ruby version >= 2.5.
 ```
 
 As it looks like the parsed gemspec is being parsed by a Ruby parser > 2.4

PR #314, opened 2023-11-24
lock aws-sdk gems to support EOL ruby versions

kumar1202:
Fixes [#313](https://github.com/hashicorp/vault-ruby/issues/313).

In order to support the Ruby version 2.3 and 2.4 in vault-ruby gem, the aws-sdk gems need to be locked to older versions where these EOL ruby versions are still supported.


averche (approved):
Thanks for finding and fixing this issue!

PR #329, opened 2024-02-27
http client's shutdown should be lock-protected

fxposter:
since `@nhp` is modified within the mutex-protected area, so all its accesses should also be mutex-protected


fxposter:
@hashicorp/vault-devex anyone?

fxposter:
@dhuckins I see that you were the last person who merged a PR to this repo, do you know who can review this?

PR #343, opened 2024-09-13
Fix encoding issue in encode_path method

TaopaiC:
Fix `encode_path` whitelist to exclude '-' from encoding


ochnerd:
@TaopaiC  would you mind adding ':' to this commit too?
We use ':' in our secrets and have the same issue.

khiav223577:
The current code escapes `~` as `%7E`, but should we be doing this? CGI.escape does not escape `~`.

PR #344, opened 2024-09-13
FIX URI path handling to remove double slashes

TaopaiC:
Fix request URI to remove double slashes.
Vault server returns 301 redirect for double slash URIs, requiring resending. Since Vault server v1.17.3, the redirect URL has changed.

For example, take key: `/cubbyhole/my-secret`:

- Before vault v1.17.2, vault-ruby sends `/v1//cubbyhole/my%2Dsecret`, and the Vault server returns `/v1/cubbyhole/my-secret`.
- After vault v1.17.3, vault-ruby sends `/v1//cubbyhole/my%2Dsecret`, and the Vault server returns `/v1/cubbyhole/my%252Dsecret`.

Thus, the behavior is not the same. Removing the double slashes prevents the redirect mechanism from being triggered.

### DIFF
--- a/lib/vault/client.rb
+++ b/lib/vault/client.rb
@@ -305,6 +305,15 @@ module Vault
       end
     end
 
+    # Removes double slashes from a path.
+    # 
+    # @param [String]
+    # 
+    # @return [String]
+    def remove_double_slash(path)
+      path.b.gsub(%r{//+}, '/')
+    end
+
     # Construct a URL from the given verb and path. If the request is a GET or
     # DELETE request, the params are assumed to be query params are are
     # converted as such using {Client#to_query_string}.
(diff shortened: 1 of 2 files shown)


ochnerd:
@TaopaiC we have a similar problem with v1.17.3. Our Problem seems to be that v1.17.3 'double-encodes' the URI.
You can see that in your request too. Your Secret `my%2Dsecret` gets encoded to `my%252Dsecret`. 
So the vault encodes the `%` to `%25`.

TaopaiC:
> @TaopaiC we have a similar problem with v1.17.3. Our Problem seems to be that v1.17.3 'double-encodes' the URI. You can see that in your request too. Your Secret `my%2Dsecret` gets encoded to `my%252Dsecret`. So the vault encodes the `%` to `%25`.

@ochnerd  We have confirmed that the issue occurred after upgrading to Go 1.22.5 in commit https://github.com/hashicorp/vault/commit/68e9f393162170ecdd3d257a42f4c8fcf3ca5aa7 .

Two possible solutions are:

- Fix the redirect double encoding issue on the Vault server (Go).
- Modify the client to remove double slashes to prevent triggering the redirect behavior (which causes the double encoding issue).

This PR attempts to implement solution 2.

PR #345, opened 2025-02-24
Un-vendor gems: upgrade `net-http-persistent` and `connection_pool`

nevans:
`net-http-persistent` and `connection_pool` were vendored, and have gotten very out-of-date.  This resets to use the upstream gems.

`connection-pool` is upgraded from 2.2.0 to 2.4+:
* 2.2.5 fixes argument forwarding for ruby 2.7+.
* 2.4.0 automatically drops all connections after fork.

`net-http-persistent` is upgraded from 3.0.0 (forked) to 4.0.2+:
* 4.0.2 fixes compatibility with `connection-pool` 2.4
* 3.1.0 supports TLS min/max and IPv6 (also in vendored fork)
* 3.1.0 fixes a memory leak in connection pooling.
* some vendored fork differences were removed:
  * use :name, :pool_size kwargs for .new

One change to the vendored fork was reimplemented: upstream does _not_ support timeout kwarg for .new, nor does `connection_pool` expose an accessor for its `timeout`.  So this code reaches into the pool and sets the `@timeout` ivar directly.

Additionally, all of the exceptions which `net-http` handles (in `Net::HTTP#transport_request`) are added to `RESCUED_EXCEPTIONS`.  Net::HTTP uses these to determine whether the socket should be closed (and whether idempotent requests should be retried).

Additionally, other changes were made so CI would pass:
* minimum ruby was set to match ruby's EOL.
* the CI matrix was updated to reflect more recent ruby and vault versions
* base64 was added as an explicit dependency, for ruby 3.4


chrisarcand:
Ah thanks for the ping. We're in a period of ownership transition here at IBM/HashiCorp so indeed, sorry for the no response here, this definitely looks like it needs doing. 

I'll get back to this by early next week and if you don't hear from me then, feel free to ping several times 😄

lib/vault/client.rb
```diff
@@ -92,7 +101,8 @@ def pool
       @lock.synchronize do
         return @nhp if @nhp
 
-        @nhp = PersistentHTTP.new("vault-ruby", nil, pool_size, pool_timeout)
+        @nhp = Net::HTTP::Persistent.new(name: "vault-ruby", pool_size:)
+        @nhp.pool.instance_variable_set(:@timeout, pool_timeout)
```
chrisarcand: This is definitely a little Hacky but but reasonable given that it doesn't appear there's any upstream method to set this.

PR #347, opened 2025-06-24
Pin dependency versions for Ruby <2.7

spectrogram:
This change pins the versions of `aws-sigv4` and `aws-eventstream` used with Ruby <2.7. AWS [recently dropped support](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sigv4/CHANGELOG.md#1120-2025-06-02) for Ruby 2.5/2.6 and because there is [no version pinning for `aws-sigv4` in the gemspec](https://github.com/hashicorp/vault-ruby/blob/master/vault.gemspec#L28) otherwise, it tries to download the latest version which is incompatible with these older versions of Ruby. 

Without this change, installations fail with something similar to the following:

```
---- Begin output of /opt/chef/embedded/bin/gem install vault -q --no-rdoc --no-ri -v "0.18.2" --source=https://www.rubygems.org ----
STDOUT:
STDERR: ERROR:  Error installing vault:
The last version of aws-eventstream (~> 1, >= 1.0.2) to support your Ruby & RubyGems was 1.3.2. Try installing it with `gem install aws-eventstream -v 1.3.2` and then running the current command again
aws-eventstream requires Ruby version >= 2.7. The current ruby version is 2.5.0.
---- End output of /opt/chef/embedded/bin/gem install vault -q --no-rdoc --no-ri -v "0.18.2" --source=https://www.rubygems.org ----
Ran /opt/chef/embedded/bin/gem install vault -q --no-rdoc --no-ri -v "0.18.2" --source=https://www.rubygems.org returned 1
```

### DIFF
--- a/vault.gemspec
+++ b/vault.gemspec
@@ -24,6 +24,9 @@ Gem::Specification.new do |spec|
   if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new("2.4.0")
     spec.add_runtime_dependency "aws-sigv4", "= 1.6.0"
     spec.add_runtime_dependency "aws-eventstream", "= 1.2.0"
+  elsif Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.7.0")
+    spec.add_runtime_dependency "aws-sigv4", "= 1.11.0"
+    spec.add_runtime_dependency "aws-eventstream", "= 1.3.2"
   else
     spec.add_runtime_dependency "aws-sigv4"
   end


khiav223577:
I’m doubtful that adding an if-condition in vault.gemspec would work as expected. When the gem is built, the .gemspec file is evaluated only once, and the resulting specification is based on the environment of the machine performing the build. 

The final gemspec uploaded to RubyGems appears to include only the resolved dependencies from the build-time environment. We can see [the vault gem](https://rubygems.org/gems/vault/versions/0.18.2) lists only a single runtime dependency: `aws-sigv4 >= 0` which reflects the line `spec.add_runtime_dependency "aws-sigv4"` in the vault.gemspec file.
image

stanhu:
> From what I can see, any conditional logic in the gemspec doesn’t appear to be reflected in the dependency data returned from that endpoint.

Right, that's what I was saying in https://github.com/hashicorp/vault-ruby/pull/347#issuecomment-3095199489.

> What do you think?

Sounds like a good plan to me.

